PR #36, opened 2016-05-13
I've finished dockerizing fbctf both in dev and prod mode :D

qazbnm456:
I've added some files in extra/ to make dockerizing more easier, and updated README as well.

Would you please have a look at it? Thanks!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM ubuntu:trusty
+MAINTAINER Boik boik@tdohacker.org
+
+ENV HOME /root
+
+ARG DOMAIN
+ARG EMAIL
+ARG MODE=dev
+ARG TYPE=self
+ARG KEY
+ARG CRT
+
+WORKDIR $HOME
+COPY . $HOME
+RUN apt-get update \
+  && apt-get install -y \
+  rsync \
+  ca-certificates \
+  && chown www-data:www-data $HOME \
+  && ./extra/provision.sh -m $MODE -c $TYPE -k $KEY -C $CRT -D $DOMAIN -e $EMAIL -s `pwd` --docker \
+  && apt-get clean \
+  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+CMD ["./extra/service_startup.sh"]
--- a/extra/lib.sh
+++ b/extra/lib.sh
@@ -110,6 +110,7 @@ function self_signed_cert() {
 function letsencrypt_cert() {
   local __email=$3
   local __domain=$4
+  local __docker=$5
 
   dl "https://dl.eff.org/certbot-auto" /usr/bin/certbot-auto
   sudo chmod a+x /usr/bin/certbot-auto
@@ -125,9 +126,21 @@ function letsencrypt_cert() {
     __mydomain=$__domain
   fi
 
-  /usr/bin/certbot-auto certonly -n --agree-tos --standalone --standalone-supported-challenges tls-sni-01 -m "$__myemail" -d "$__mydomain"
-  sudo ln -s "/etc/letsencrypt/live/$__mydomain/cert.pem" "$1" || true
-  sudo ln -s "/etc/letsencrypt/live/$__mydomain/privkey.pem" "$2" || true
+  if [[ $__docker = true ]]; then
+    cat <<- EOF > /root/tmp/certbot.sh
+		#!/bin/bash
+		if [[ ! ( -d /etc/letsencrypt && "\$(ls -A /etc/letsencrypt)" ) ]]; then
+		    /usr/bin/certbot-auto certonly -n --agree-tos --standalone --standalone-supported-challenges tls-sni-01 -m "$__myemail" -d "$__mydomain"
+		fi
+		sudo ln -sf "/etc/letsencrypt/live/$__mydomain/cert.pem" "$1"
+		sudo ln -sf "/etc/letsencrypt/live/$__mydomain/privkey.pem" "$2"
+EOF
+    sudo chmod +x /root/tmp/certbot.sh
+  else
+    /usr/bin/certbot-auto certonly -n --agree-tos --standalone --standalone-supported-challenges tls-sni-01 -m "$__myemail" -d "$__mydomain"
+    sudo ln -s "/etc/letsencrypt/live/$__mydomain/cert.pem" "$1" || true
+    sudo ln -s "/etc/letsencrypt/live/$__mydomain/privkey.pem" "$2" || true
+  fi
 }
 
 function own_cert() {
@@ -146,6 +159,7 @@ function install_nginx() {
   local __certs=$3
   local __email=$4
   local __domain=$5
+  local __docker=$6
 
   local __certs_path="/etc/nginx/certs"
 
@@ -167,7 +181,10 @@ function install_nginx() {
         own_cert "$__cert" "$__key"
       ;;
       certbot)
-        letsencrypt_cert "$__cert" "$__key" "$__email" "$__domain"
+        if [[ $__docker = true ]]; then
+          self_signed_cert "$__cert" "$__key"
+        fi
+        letsencrypt_cert "$__cert" "$__key" "$__email" "$__domain" "$__docker"
       ;;
       *)
         error_log "Unrecognized type of certificate"
--- a/extra/service_startup.sh
+++ b/extra/service_startup.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -e
+
+if [[ -e /root/tmp/certbot.sh ]]; then
+    /bin/bash /root/tmp/certbot.sh
+fi
+
+service nginx restart
+service mysql start
+service hhvm start
+
+while true; do
+    if [[ -e /var/log/nginx/access.log ]]; then
+        exec tail -F /var/log/nginx/access.log
+    else
+        exec sleep 10
+    fi
+done
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.git
+.gitignore
+LICENSE
+screencapture.gif
--- a/README.md
+++ b/README.md
@@ -120,6 +120,45 @@ Note that the unison script will not sync NPM dependencies, so if you ever need 
 
 This step is not necessary if all development is done on the VM.
 
+# Using Docker
+
+[Dockerfile](Dockerfile) is provided, you can use docker to deploy fbctf to both development and production.
+
+### Development
+
+This build command will provision fbctf in dev mode in docker:
+
+`docker build --build-arg MODE=dev -t="fbctf_in_dev" .`
+
+
+Run command:
+
+`docker run -p 80:80 -p 443:443 fbctf_in_dev`
+
+Once you've started the container, go to the URL/IP of the server. Click the "Login" link at the top right, enter the default admin credentials, and you'll be redirected to the admin page.
+
+### Production
+
+This build command will provision fbctf in prod mode in docker. Once you've started the container, Let's Encrypt will take YOUR_DOMAIN to obtain a trusted certificate at zero cost:
+
+`docker build --build-arg MODE=prod --build-arg DOMAIN=test.mydomain.com --build-arg EMAIL=myemail@mydomain.com --build-arg TYPE=certbot -t="fbctf_in_prod" .`
+
+Run command:
+
+`docker run -p 80:80 -p 443:443 -v /etc/letsencrypt:/etc/letsencrypt fbctf_in_prod`
+
+*Note 1: Because this is a production environment, the password will be randomly generated when the provision script finishes. Make sure to watch the very end of the provision script while image is building, as the password will be printed out. It will not be stored elsewhere, so either keep track of it or change it. In order to change the password, run the following command in container:*
+
+```
+set_password new_password ctf ctf fbctf /root
+```
+
+*This will set the password to 'new_password', assuming the database user/password is ctf/ctf and the database name is fbctf (these are the defaults).*
+
+*Note 2: You'll have to mount /etc/letsencrypt as a volume to persist the certs. ex: `docker run -v /etc/letsencrypt:/etc/letsencrypt ...`*
+
+You can go to the URL/IP of the server, click the "Login" link at the top right, enter the admin credentials, and you'll be redirected to the admin page.
+
 ## Reporting an Issue
 
 First, ensure the issue was not already reported by doing a search. If you cannot find an existing issue, create a new issue. Make the title and description as clear as possible, and include a test case or screenshot to reproduce or illustrate the problem if possible.
--- a/extra/provision.sh
+++ b/extra/provision.sh
@@ -84,7 +84,7 @@ function usage() {
   printf "\t%s -m prod -c own -k /etc/certs/my.key -C /etc/certs/cert.crt -s /home/foobar/fbctf -d /var/fbctf\n" "${0}"
 }
 
-ARGS=$(getopt -n "$0" -o hm:c:k:C:D:e:s:d: -l "help,mode:,cert:,keyfile:,certfile:,domain:,email:,code:,destination:,update" -- "$@")
+ARGS=$(getopt -n "$0" -o hm:c:k:C:D:e:s:d: -l "help,mode:,cert:,keyfile:,certfile:,domain:,email:,code:,destination:,update,docker" -- "$@")
 
 eval set -- "$ARGS"
 
@@ -142,6 +142,10 @@ while true; do
       UPDATE=true
       shift
       ;;
+    --docker)
+      DOCKER=true
+      shift
+      ;;
     --)
       shift
       break
@@ -186,6 +190,9 @@ fi
 # Ascii art is always appreciated
 set_motd "$CTF_PATH"
 
+# Some Ubuntu distros don't come with curl installed
+package curl
+
 # Repos to be added in dev mode
 if [[ "$MODE" == "dev" ]]; then
     repo_mycli
@@ -193,9 +200,6 @@ fi
 
 # We only run this once so provisioning is faster
 sudo apt-get update
-
-# Some Ubuntu distros don't come with curl installed
-package curl
 
 # Some people need this language pack installed or HHVM will report errors
 package language-pack-en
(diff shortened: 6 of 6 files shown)


facebook-github-bot-8:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

qazbnm456:
I've alredy signed it. Thanks.

facebook-github-bot-4:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

pdxjohnny:
lol I was about to do this! thanks!!

AlexGaspar:
I also started something, but I took a different path, I ended up not using `./extra/provision.sh`, since half the things it runs doesn't make sense for docker (emacs, htop, mycli, unison, ...). What do you think?

I also think it would be better to split the processes into multiple containers (MySQL, memcached, hhvm/nginx). Having multiple processes per container is a bit against the docker philosophy isn't it ? :)

qazbnm456:
@AlexGaspar I agree with you, there are lots of stuffs that docker actually doesn't need them to be installed. Right now, I'm just making an attempt to modify less changes as possible as I can to let docker run successfully. Maybe we could try to move dockerizing to another provision script eventually.

Spliting processes into multiple containers is indeed a better way, we can do this after things mentioned above have been resolved. Or do you have any progress on it? Thanks for your suggestions btw :)

Hoilam:
why?

gsingh93:
Thanks for the PR! We'll review it over the next few days.

I definitely agree some tools aren't necessary for prod mode (i.e. emacs, htop, mycli, unison). We'll modify the provision script to install these only in dev mode.

We're fine with either the single container or multiple container approach.

AlexGaspar:
@qazbnm456 Yep I started this : https://github.com/AlexGaspar/docker-fbctf if you find value in what I've done we could perhaps merge our 2 things?

I've splinted the resources into multi containers, but I didn't include the SSL termination, I don't think it should be the responsibility of the fbctf container. In the other hand, fbctf force the session's cookie to `Secure`, making it impossible to log in, in a no-SSL environment.

MikeMichel:
@AlexGaspar i agree to both, it should be splitted and ssl termination should be the job of the orchestration platform which often comes with it's own loadbalancer. in my case i want to make fbctf ready for sloppy.io but right now it's not possible. i changed the nginx conf to http but login is not working.

RW34:
hey guys i am new in this  and i am arabian so can you give me some help

qazbnm456:
@AlexGaspar Neat, I'll look into it when I/m available, many thanks :)

@AlexGaspar @MikeMichel Due to [https://github.com/facebook/fbctf/blob/5d75a0a5b71d724ff71ac301e99ad5ed99a7947b/src/SessionUtils.php#L6](https://github.com/facebook/fbctf/blob/5d75a0a5b71d724ff71ac301e99ad5ed99a7947b/src/SessionUtils.php#L6), I'll reserve my approach about ssl termination included container, which may help those who are not familiar with ctf get involved more easily. Thanks for your suggestions, I'm really appreciate. :)

qazbnm456:
@RW34 What kind of help do you need? :)

gsingh93:
While I completely understand why people would want to have their own control over the way SSL is implemented, we wanted to make the platform as simple as possible for someone to setup and use. We assume that people who want more advanced setups can tweak the platform as necessary, while people with less experience can use it as is and still have a secure setup.

Jovonni:
YOUR DOCKER CONTAINER ACTUALLY HAS FLAWS IN IT. It would not build, and hangs up at the running MYSQL phase of the end of the process

qazbnm456:
@Jovonni Do you mean you get stuck in your building process on ``` Step 10 : RUN ./extra/provision.sh dev `pwd` ```? If that's the case, you can just wait for it, because docker is still processing.

Jovonni:
@qazbnm456 No, I actually let it run for almost an hour. I also tried it on two different machines running docker, it still hangs on the same part. I don't know what it is. Have you not encountered this?

AlexGaspar:
@Jovonni: I made [this](https://github.com/AlexGaspar/docker-fbctf) does that work for you?

qazbnm456:
@Jovonni Weird, I've not encountered this... Following is my recording of building process:

[asciicast](https://asciinema.org/a/45725?autoplay=1)

By the way, could you show me your recording as well which can possibly let me dive into solving potential problems. Thank you for your feedback.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+COPY extra/https_fbctf.conf /etc/nginx/sites-available/https_fbctf.conf
+RUN sed -i 's/domain.com/'$DOMAIN'/g' /etc/nginx/sites-available/http_fbctf.conf \
+  && sed -i 's/domain.com/'$DOMAIN'/g' /etc/nginx/sites-available/https_fbctf.conf
+
+RUN mkdir -p /etc/my_init.d
+COPY extra/setup_lets_encrypt.sh /etc/my_init.d/01_setup_lets_encrypt.sh
+COPY extra/service_startup.sh /etc/my_init.d/02_service_startup.sh
+COPY extra/import_empty_db.sh /etc/my_init.d/03_import_empty_db.sh
+RUN [ "$MODE" = "dev" ] && rm /etc/my_init.d/01_setup_lets_encrypt.sh || true
+RUN [ "$MODE" = "prod" ] && sed -i 's/domain.com/'$DOMAIN'/g' /etc/my_init.d/01_setup_lets_encrypt.sh || true
```
aaronjwood: I think many people will be terminating TLS at their reverse proxy. Is it possible to skip all of the TLS stuff inside of the container?

Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM qazbnm456/fbctf_base:latest
```
aaronjwood: Is there a big benefit to having this base image? Can the official Ubuntu 14.04.x image be used instead?

Dockerfile
qazbnm456: As @gsingh93 said, this Dockerfile aims to be friendly to newbies, which means advanced hackers are supposed to tweak the platform as necessary. Thus, no modification will be made currently.

Dockerfile
qazbnm456: The image "qazbnm456/fbctf_base" is just doing pre-install some softwares to make building process more quickly, and its source code is also available on github. If you want to replace it, use phusion/baseimage:latest instead. The reason is that qazbnm456/fbctf_base is built from that image, and some provided scripts are also dependent on that , too.

aaronjwood: Ah, I see. I was just curious since the documentation states only Ubuntu 14.04 is needed.

koenpunt: Still true I guess, but this custom image probably has all the packages installed that would be installed by the setup script, so that doesn't have to happen for people using this container image

extra/config.sh
```diff
@@ -0,0 +1,72 @@
...
+# /etc/letsencrypt.sh/config.sh                        #
+# ${PWD}/config.sh (in current working-directory)      #
+#                                                      #
+# Default values of this config are in comments        #
+########################################################
+
+# Path to certificate authority (default: https://acme-v01.api.letsencrypt.org/directory)
+
+# Path to license agreement (default: https://letsencrypt.org/documents/LE-SA-v1.0.1-July-27-2015.pdf)
+#LICENSE="https://letsencrypt.org/documents/LE-SA-v1.0.1-July-27-2015.pdf"
```
gsingh93: Can you remove all lines that are commenting out unused variables?

extra/setup_lets_encrypt.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -e
+
+curl -h >/dev/null 2>&1 || { echo >&2 "Curl is needed but I cannot find that. Aborting."; exit 1; }
+openssl -h >/dev/null 2>&1 || { echo >&2 "Openssl is needed but I cannot find that. Aborting."; exit 1; }
+
+PWD="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $PWD \
+    && git clone https://github.com/lukas2511/letsencrypt.sh.git \
```
gsingh93: These `&&`s look kind of messy. You're already using `set -e` so I don't see why you can't just put each command on it's own line.

gsingh93: I also realized this isn't the official let's encrypt package. Any reason to use this wrapper instead of the original package?

gsingh93:
I'll be looking at this and testing it locally once I get #105 merged, at which point this should be rebased on top of that.

Overall it looks good other than the minor things I already commented on. I'd also like to not duplicate the `import_empty_db` functionality, so it'd be nice if it was only defined once.

We're going to have to host our own version of your `fbctf_base` image before merging, I'll take care of setting that up.

extra/config.sh
qazbnm456: Sure!

extra/setup_lets_encrypt.sh
qazbnm456: Okay, I'll remove all `&&` since they are indeed unnecessary.

Compared to Official letsencrypt client, which needs python and other dependent softwares, it's more easily to get used to letsencrypt.sh, which only needs curl and openssl.

qazbnm456:
@gsingh93 Okay, I'll await your work and rebase this PR while yours get merged. Then, I'll do further modification if it need to.

javuto:
@qazbnm456 can you rebase this? I'd like to merge this awesome PR soon :)

qazbnm456:
@javuto Sure! I will rebase it a.s.a.p. :)

qazbnm456:
@javuto I've finished the rebase. :)

javuto:
Looks good to me, @gsingh93?

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN ./extra/provision.sh -m $MODE -s `pwd`
+RUN [ "$MODE" = "prod" ] && rm /etc/nginx/sites-enabled/fbctf.conf || true
+
+COPY extra/http_fbctf.conf /etc/nginx/sites-available/http_fbctf.conf
+COPY extra/https_fbctf.conf /etc/nginx/sites-available/https_fbctf.conf
+RUN sed -i 's/domain.com/'$DOMAIN'/g' /etc/nginx/sites-available/http_fbctf.conf \
+  && sed -i 's/domain.com/'$DOMAIN'/g' /etc/nginx/sites-available/https_fbctf.conf
+
+RUN mkdir -p /etc/my_init.d
+COPY extra/setup_lets_encrypt.sh /etc/my_init.d/01_setup_lets_encrypt.sh
```
gsingh93: You can just copy if it's in prod mode, you don't have to copy and delete later.

Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM phusion/baseimage:latest
```
gsingh93: I understand why you want to use this instead of `ubuntu:latest`, but this allows external people to put arbitrary code on our docker instances. I'd feel safer if we just used `ubuntu:latest` and trusted canonical.

We could alternatively build and host this baseimage ourselves on our dockerhub account, but that's more work. We can always do that in the future though, so for now I think this should be changed to `ubuntu:latest`.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+WORKDIR $HOME
+RUN apt-get update \
+  && apt-get install -y rsync
+ADD . $HOME
+RUN chown root:root $HOME
+
+RUN ./extra/provision.sh -m $MODE -s `pwd`
+RUN [ "$MODE" = "prod" ] && rm /etc/nginx/sites-enabled/fbctf.conf || true
+
+COPY extra/http_fbctf.conf /etc/nginx/sites-available/http_fbctf.conf
```
gsingh93: We already provide an nginx configuration, is this specifically for letsencrypt? We have built in letsencrypt support now, you can just use that.

gsingh93:
I think this will break when you run with `MODE=prod`. Also, why don't you just use the built in letsencrypt stuff now for provisioning? See the README for how to use letencrypt.

qazbnm456:
@javuto @gsingh93 Thanks for your comments! I'll revise my codes a.s.a.p.

qazbnm456:
@gsingh93 I've made some changes. Does it look good right now?

extra/lib.sh
```diff
@@ -167,7 +177,10 @@ function install_nginx() {
         own_cert "$__cert" "$__key"
       ;;
       certbot)
-        letsencrypt_cert "$__cert" "$__key" "$__email" "$__domain"
+        if [[ $__docker = true ]]; then
```
gsingh93: Why do we need to do this?

gsingh93:
Looking much better. I'm assuming you're creating certbot.sh to renew the cert on load it expires? Or is it for a different reason?

qazbnm456:
certbot.sh is actually doing the same thing. What resides in certbot.sh is something like:
```bash
#!/bin/bash
/usr/bin/certbot-auto certonly -n --agree-tos --standalone --standalone-supported-challenges tls-sni-01 -m "xxx@xxx.com" -d "xxx.com"
sudo ln -sf "/etc/letsencrypt/live/xxx.com/cert.pem" "/etc/nginx/certs/fbctf.crt"
sudo ln -sf "/etc/letsencrypt/live/xxx.com/privkey.pem" "/etc/nginx/certs/fbctf.key"
```
The reason for doing this is while we are building the fbctf image, there will no available internet connections for **certbot-auto** to get our certs. Thus, I've added the docker flag as a hint to tell **provision.sh** to postpone the process of getting certs, and "writing those scripts into another certbot.sh" is the countermeasure I choose to implement.

As you can see that in **service_startup.sh**, it will check if there is a **certbot.sh** in ~/tmp/. If there is one, it will start the script to get required certs.

extra/lib.sh
qazbnm456: I've explained it in detailed below.

gsingh93:
Ok, that makes sense, but the only thing I'm worried about now is that if the cert is created when you run the docker container, running the container 5 times will generate 5 different certs for the same domain and then you'll hit the letsencrypt rate limit. Any thoughts on that?

qazbnm456:
There are actually many ways users can do to prevent recreating certs. All they must to do is copy those certs in container to host, then they can choose to rebuild image with type=own , or overwrite the CMD and mount their certs into container.

Therefore, I think it depends on how users interact to their containers. Docker is just another easy way for beginners to run an instance, and they can tune it by themselves. But a line of notice is probably needed :)

How do you think, @gsingh93 ?

extra/service_startup.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -e
+
+if [[ -e /root/tmp/certbot.sh ]]; then
+    /bin/bash /root/tmp/certbot.sh
```
javuto: Once we execute this, we can clear the file, so there is no risk on generating a new certificate again.

extra/lib.sh
```diff
@@ -125,9 +126,17 @@ function letsencrypt_cert() {
     __mydomain=$__domain
   fi
 
-  /usr/bin/certbot-auto certonly -n --agree-tos --standalone --standalone-supported-challenges tls-sni-01 -m "$__myemail" -d "$__mydomain"
-  sudo ln -s "/etc/letsencrypt/live/$__mydomain/cert.pem" "$1" || true
-  sudo ln -s "/etc/letsencrypt/live/$__mydomain/privkey.pem" "$2" || true
+  if [[ $__docker = true ]]; then
+    echo "#!/bin/bash" >> /root/tmp/certbot.sh
```
javuto: This is nasty, why don't we have a ```certbot_helper.sh``` already created distributed by our platform, and we either run it or copy it to the ```/tmp``` location?

extra/service_startup.sh
javuto: Nevermind my comment, I didn't realize the file will still be there inside the container. The only way to handle this without taking all the certificates will be to handle that outside of the container and just copy the files...

extra/lib.sh
qazbnm456: I was just making an attempt to build the fbctf image in prod successfully. I know the code above is nasty, and maybe you can make me the `certbot_helper.sh`, which I can use it to let the process of dockerizing more neatly.

gsingh93:
Can we mount /etc/letsencrypt as a volume, so the data there persists?

qazbnm456:
@gsingh93 Yes, we can. However, `certbot.sh` will still be triggered when container restart again.

 I think I can make `service_startup.sh` check if there are already some certs in /etc/letsencrypt. If so, it will skip `/usr/bin/certbot-auto certonly -n --agree-tos --standalone --standalone-supported-challenges tls-sni-01 -m "xxx@xxx.com" -d "xxx.com"` and just link those certs to /etc/nginx/certs.

qazbnm456:
@javuto @gsingh93 Okay, I've made some changes. This time, If we mount /etc/letsencrypt as a volume, it will call `certbot-auto` to get required certs only at the first time. When we restart the container and mount the same volume, it will not regenerate the certs cause there are certs in /etc/letsencrypt. How do you think? Is it much better now?

extra/lib.sh
gsingh93: Could you at least make this a heredoc?

http://tldp.org/LDP/abs/html/here-docs.html

gsingh93:
Other than my comment about the heredoc above, this looks good now.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+ARG KEY
+ARG CRT
+
+WORKDIR $HOME
+RUN apt-get update \
+  && apt-get install -y rsync ca-certificates
+ADD . $HOME
+RUN chown www-data:www-data $HOME
+RUN ./extra/provision.sh -m $MODE -c $TYPE -k $KEY -C $CRT -D $DOMAIN -e $EMAIL -s `pwd` --docker
+RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
gsingh93: I'm not docker expert, but I've heard it's better practice to put lines like this in the same `RUN` command as the ones where the files were added (the `apt-get update` command). The reason is even though you delete them before the final image is created, docker has already created a FS layer that includes the files, so the image size still increases.

gsingh93:
Also, maybe the README should contain an example `docker up` command, or maybe a docker-compose.yml file showing that using a persistent volume for certificates is recommended.

javuto:
Kickass PR @qazbnm456, @gsingh93 merge at will!

gsingh93:
Thanks @qazbnm456!

qazbnm456:
@javuto @gsingh93 You guys rock! Many thanks :)

PR #83, opened 2016-05-18
Added envvar options

5003:
provider: VB


gsingh93:
Thanks for the PR!

Instead of choosing the max, I'd prefer to have the defaults set to 4/4096, and then if those environment variables are set at all, the environment variables should be used, regardless of whether they're larger or smaller than the defaults.

gsingh93:
Closing due to inactivity.

PR #107, opened 2016-05-26
first_capture does not return the correct result

tinduong94:
The scoreboard first_capture always return the first team in the order of team name instead of the order of captured time.

### DIFF
--- a/src/data/country-data.php
+++ b/src/data/country-data.php
@@ -76,7 +76,11 @@ class CountryDataController extends DataController {
       }
 
       // Who is the first owner of this level
-      $owner = ($completed_by) ? $completed_by[0] : 'Uncaptured';
+      if ($completed_level) {
+        $owner = await Team::getOwnerLevel($level->getId());
+        $owner = $owner->getName();
+      } else
+        $owner = 'Uncaptured';
       $country_data = (object) array(
         'level_id'    => $level->getId(),
         'title'       => $level->getTitle(),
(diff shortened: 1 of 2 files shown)


javuto:
Nice job @tinduong94 ! I will be happy to merge this after one little comment I added to the code

src/data/country-data.php
```diff
@@ -76,7 +76,11 @@ class CountryDataController extends DataController {
       }
 
       // Who is the first owner of this level
-      $owner = ($completed_by) ? $completed_by[0] : 'Uncaptured';
+      if ($completed_level) {
+        $owner = await Team::getOwnerLevel($level->getId());
+        $owner = $owner->getName();
+      } else
```
javuto: Can you add braces around the else clause? Thanks!

src/models/Team.php
```diff
@@ -490,6 +490,17 @@ public static function regenerateHash(string $password_hash): bool {
     return intval(idx($result->mapRows()[0], 'COUNT(*)'));
   }
 
+  public static async function getOwnerLevel(
```
javuto: It would follow better convention to call the method ```genFirstCapture```

tinduong94:
Check again @javuto :)
Thanks !

javuto:
@tinduong94 perfect! Squash the commits and you are good to go!

javuto:
@tinduong94 nevermind, you can actually do that now from the merge. Good job! :+1:

PR #112, opened 2016-05-27
Fix typos in README.md

gautamg795:
'English' spelled incorrectly in readme!


javuto:
Good catch @gautamg795

PR #115, opened 2016-05-31
Production deployment using letsencrypt certs

javuto:
For production deployments, you can now use letsencrypt to create certificates. It uses certbot

### DIFF
--- a/extra/lib.sh
+++ b/extra/lib.sh
@@ -85,11 +85,11 @@ function run_grunt() {
 
 function self_signed_cert() {
   local __csr="/etc/nginx/certs/dev.csr"
-  local __cert="/etc/nginx/certs/dev.crt"
-  local __key="/etc/nginx/certs/dev.key"
+  local __devcert=$1
+  local __devkey=$2
 
-  sudo openssl req -nodes -newkey rsa:2048 -keyout "$__key" -out "$__csr" -subj "/O=Facebook CTF"
-  sudo openssl x509 -req -days 365 -in "$__csr" -signkey "$__key" -out "$__cert"
+  sudo openssl req -nodes -newkey rsa:2048 -keyout "$__devkey" -out "$__csr" -subj "/O=Facebook CTF"
+  sudo openssl x509 -req -days 365 -in "$__csr" -signkey "$__devkey" -out "$__devcert"
 }
 
 function letsencrypt_cert() {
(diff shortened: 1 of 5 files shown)


gsingh93:
As discussed offline, let's also modify the provision script to take optional flags to choose whether to use letsencrypt or not, and also add some flags to make sure the provision script doesn't need user input during execution so it can be completely automated.

extra/lib.sh
```diff
@@ -7,6 +7,21 @@ function log() {
   echo "[+] $1"
 }
 
+function error() {
+  echo "[!] $1"
```
gsingh93: nit: make this red

extra/lib.sh
```diff
@@ -72,32 +87,94 @@ function run_grunt() {
...
+    self_signed_cert "$__cert" "$__key"
   elif [[ $__mode = "prod" ]]; then
-    __cert="/etc/nginx/certs/fbctf.csr"
-    __key="/etc/nginx/certs/fbctf.key"
-    read -p ' -> SSL Certificate file location? ' __mycert
-    read -p ' -> SSL Key Certificate file location? ' __mykey
-    sudo cp "$__mycert" "$__cert"
-    sudo cp "$__mykey" "$__key"
+    local __cert="$__certs_path/fbctf.crt"
+    local __key="__certs_path/fbctf.key"
```
gsingh93: missing a $ here before __certs

extra/provision.sh
```diff
@@ -1,9 +1,34 @@
...
+#   dev    Provision will run in development mode. Certificate will be self-signed.
+#   prod   Provision will run in production mode.
+#   self   Provision will use a self-signed SSL certificate that will be generated.
+#   own    Provision will use the SSL certificate provided by the user.
+#   certbot Provision will generate a SSL certificate using letsencrypt/certbot. More info here: https://certbot.eff.org/
+#
+# Optional Parameters:
+#   -k PATH, --keyfile PATH      Path to supplied SSL key file.
+#   -C PATH, --certfile PATH     Path to supplied SSL certificate pem file.
+#   -d DOMAIN, --domain DOMAIN   Domain for the SSL certificate to be generated using letsencrypt.
```
gsingh93: Let's make this capital `D` since we have capital `C` here as well. For source and dest we can use lower case letters (`s` and `d`).

extra/provision.sh
```diff
@@ -1,9 +1,34 @@
...
-# Usage: ./provision.sh [dev | prod] [path_to_code] [path_to_destination]
+# Usage: usage [-h|--help] [PARAMETER [ARGUMENT]] [PARAMETER [ARGUMENT]] ...
 #
+# Parameters:
+#   -h, --help            Shows this help message and exit.
+#   -m MODE, --mode MODE  Mode of operation. Default value is dev
+#   -c TYPE, --cert TYPE  Type of certificate to use. Default value is self
+#
+# Arguments:
+#   dev    Provision will run in development mode. Certificate will be self-signed.
```
gsingh93: You should say that these three arguments correspond to MODE and the next two correspond to TYPE.

extra/provision.sh
```diff
@@ -1,9 +1,34 @@
...
+#   prod   Provision will run in production mode.
+#   self   Provision will use a self-signed SSL certificate that will be generated.
+#   own    Provision will use the SSL certificate provided by the user.
+#   certbot Provision will generate a SSL certificate using letsencrypt/certbot. More info here: https://certbot.eff.org/
+#
+# Optional Parameters:
+#   -k PATH, --keyfile PATH      Path to supplied SSL key file.
+#   -C PATH, --certfile PATH     Path to supplied SSL certificate pem file.
+#   -d DOMAIN, --domain DOMAIN   Domain for the SSL certificate to be generated using letsencrypt.
+#   -m EMAIL, --email EMAIL      Domain for the SSL certificate to be generated using letsencrypt.
```
gsingh93: You're using `m` for mode as well. What about `e` here?

extra/lib.sh
```diff
@@ -7,8 +7,16 @@ function log() {
   echo "[+] $1"
 }
 
-function error() {
-  echo "[!] $1" 
+function error_log() {
+  RED='\033[0;31m'
+  NORMAL='\033[0m'
+  echo "${RED} [!] $1 ${NC}"
```
gsingh93: Do you want to use `${NC}` here or `${NORMAL}`?

extra/lib.sh
```diff
@@ -7,8 +7,16 @@ function log() {
...
+}
+
+function ok_log() {
+  GREEN='\033[0;32m'
+  NORMAL='\033[0m'
+  echo "${GREEN} [+] $1 ${NC}"
```
gsingh93: Same as above.

extra/provision.sh
```diff
@@ -12,11 +39,116 @@ DB="fbctf"
...
+      ;;
+    *)
+      usage
+      exit 1
+      ;;
+  esac
+done
 
 echo "[+] Provisioning in $MODE mode"
+echo "[+] Using $TYPE certificate"
```
gsingh93: Can you print out the `code` and `destination` paths here too? It will help when debugging user bug reports.

Issue #78, opened 2016-05-17
Allow longer/shorter game times


landmanmanc:
Feature request : custom gametime longer then 24 hours.

colbyjoe1919:
A day here is a 1000 years in spirit prison.  So your penance might take a
minute.

gsingh93:
Yea, I was planning on changing this. Thanks for the issue!

landmanmanc:
Would be nice to set custom times, and not just 12/24/48 hours.

javuto:
Implemented in https://github.com/facebook/fbctf/pull/118

PR #118, opened 2016-06-02
Configuration option for arbitrary game duration added

markszabo:
As requested in #78 the possibility to select arbitrary game duration is implemented:
selection_001
TODO: if a duration longer than 99 hours is selected, then the Gameclock overlaps:
selection_002

### DIFF
--- a/database/schema.sql
+++ b/database/schema.sql
@@ -184,7 +184,8 @@ CREATE TABLE `configuration` (
 LOCK TABLES `configuration` WRITE;
 INSERT INTO `configuration` (field, value, description) VALUES("game", "0", "(Boolean) Game is ongoing");
 INSERT INTO `configuration` (field, value, description) VALUES("next_game", "0", "(Date) Next game to happen");
-INSERT INTO `configuration` (field, value, description) VALUES("game_duration", "10800", "(Integer) Duration of game in seconds");
+INSERT INTO `configuration` (field, value, description) VALUES("game_duration_value", "3", "(Integer) Value of the duration of the game");
+INSERT INTO `configuration` (field, value, description) VALUES("game_duration_unit", "h", "(Character) Unit of the duration of the game");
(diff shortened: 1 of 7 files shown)


gsingh93:
Thanks for the PR! While we do want this feature, we probably need to figure out how to fix the overlap before we do this. We probably just need to space out the `:` a bit more.

markszabo:
I fixed the overlap, now it looks like this:
Normal operation below 24 hours is not changed:
selection_004
Above 24 hours the milliseconds are gone, and we have days instead:
selection_003
selection_002
It also works above 99 days:
selection_001
What do you think?

javuto:
Let's get this merged! Can you rebase with current master and update this? I can merge it right away. Thanks again for the PR @markszabo!

markszabo:
I just did the rebase and resolved the conflicts. Sorry for the long reply, I was AFK.

javuto:
Looks good, any concerns here @gsingh93?

javuto:
On second thoughts, we need a way to make a game to last 1 day, 3 hours and 17 minutes. Probably we can integrate it here. Thoughts @markszabo?

markszabo:
Well it's not too complicated to integrate, but do we really need this? I mean people will usually start games for 5 hour, for 2 days or for 30 days. And if they really want to start one for 1 day and 3 hours, they can still start it for 27 hours. But if you think it's really needed, I can do it.

gsingh93:
@markszabo your PR is a bit messed up because it includes commits from other people. Can you remove the extra commits from the PR? Let me know if you need help.

I think we can leave off the minutes stuff for this PR. I can see it being useful if we for some reason want to change the end time while the game is running (i.e. we'd set a new end time that would be 1 hour and 13 minutes from now), but I don't think we support changing the end time anyway, so it's fine for now.

markszabo:
I did the rebase again, and those commits disappeared. Is it okay now?

PR #119, opened 2016-06-03
Added Ubuntu/trusty64 VM.box URL

AshutoshKumarAnand:


gsingh93:
Thanks for the PR! I don't think this parameter is necessary, is there any reason you want to add it? If not, I'll be closing this PR.

javuto:
Agree with @gsingh93, not sure what is the benefit of this.

AshutoshKumarAnand:
The person who is first time installing on their Ubuntu machine were
getting errors.

Also directly adding Ubuntu trusty 64 is not working. So this would help
any user to directly install fbctf without any problem.

gsingh93:
Can you elaborate on what the errors are and the exact steps you're doing? I bet there's a better fix.

PR #131, opened 2016-06-11
add user / pass for VM

mtmcgrew:
This adds the username and password for the vagrant VM to the docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ cd fbctf
 vagrant up
 ```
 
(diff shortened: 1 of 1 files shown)


gsingh93:
The username and password are in the paragraph below the one you modified. Do you think we should move it?

mtmcgrew:
admin / password? That for the web ui, not the vm, right?

gsingh93:
Yea, that's for the web UI, I didn't realize you were talking about the VM. Are you not using `vagrant` from the command line? `vagrant ssh` should log you in without the need for a password.

mtmcgrew:
Ahhh ok, that makes sense. I'll change the wording to that. I don't think that's obvious to someone un-familiar with vagrant. Someone asked me today at a ctf how to log into the vm.

gsingh93:
That's a good point, I'll add the note about `vagrant ssh` to the README. Are you fine with me closing this PR?

PR #144, opened 2016-06-21
Fixed scoring race condition

gsingh93:
There was a race condition which would allow people to score twice by submitting two requests concurrently. Teams could also request two hints even if they only had points for one hint, because of this race condition. This PR should fix both issues.

### DIFF
--- a/src/models/Level.php
+++ b/src/models/Level.php
@@ -654,38 +654,72 @@ class Level extends Model {
     );
   }
 
+  private static async function withLock(
+    string $lock_name,
+    int $team_id,
+    (function(): Awaitable<mixed>) $thunk,
+  ): Awaitable<mixed> {
+    $lock_name = sprintf('%s/%s_%d', sys_get_temp_dir(), $lock_name, $team_id);
+    $lock = fopen($lock_name, 'w');
+
+    if ($lock === false) {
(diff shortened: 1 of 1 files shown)


src/models/Level.php
```diff
@@ -654,106 +654,156 @@ private static function levelFromRow(Map<string, string> $row): Level {
     );
   }
 
+  private static async function withLock(
+    string $lock_name,
+    int $team_id,
+    (function(): Awaitable<mixed>) $thunk,
+  ): Awaitable<mixed> {
+    $lock_name = sprintf('%s/%s_%d', sys_get_temp_dir(), $lock_name, $team_id);
```
javuto: Do we want to provide timestamp here somewhere?

javuto:
Looks pretty good to me! I added one comment just in case we want to make sure there is a isolation for the same team

src/models/Level.php
gsingh93: I was thinking of having each team reuse their lock file for the duration of the CTF. The benefit of this is we don't need to create a new file on each request. I don't see a downside of this reuse.

javuto: Fair enough. That is a good solution. Push push! ![Approved] (http://a.fod4.com/misc/rambo%20original.gif)

PR #150, opened 2016-06-27
Adding exporting/importing capabilities

javuto:
This is a quite substantial diff so it will require testing and review. Also there are errors in ```hh_client```

### DIFF
--- a/src/controllers/exporter/ExporterController.php
+++ b/src/controllers/exporter/ExporterController.php
@@ -1,0 +1,99 @@
+<?hh // strict
+
+abstract class ExporterController {
+  abstract public function genData(string $type): Awaitable<mixed>;
+
+  public async function genTeams(): Awaitable<mixed> {
+		$all_teams_data = array();
+  	$all_teams = \HH\Asio\join(Team::genAllTeams());
+
+  	foreach ($all_teams as $team) {
+  		$team_data = \HH\Asio\join(Team::genTeamData($team->getId()));
+  		$one_team = array(
+  			'name' => $team->getName(),
+  			'active' => $team->getActive(),
+  			'admin' => $team->getAdmin(),
+  			'protected' => $team->getProtected(),
+  			'visible' => $team->getVisible(),
+  			'password_hash' => $team->getPasswordHash(),
+  			'points' => $team->getPoints(),
+  			'logo' => $team->getLogo(),
+  			'data' => $team_data
+  		);
+  		array_push($all_teams_data, $one_team);
+  	}
+  	return $all_teams_data;
+  }
+
+  public async function genLevels(): Awaitable<mixed> {
+  	$all_levels_data = array();
+  	$all_levels = \HH\Asio\join(Level::genAllLevels());
+
+  	foreach ($all_levels as $level) {
+      $entity = \HH\Asio\join(Country::gen($level->getEntityId()));
+      $category = \HH\Asio\join(Category::genSingleCategory($level->getCategoryId()));
+      $one_level = array(
+        'type' => $level->getType(),
+        'title' => $level->getTitle(),
+        'active' => $level->getActive(),
+        'description' => $level->getDescription(),
+        'entity_name' => $entity->getName(),
+        'category' => $category->getCategory(),
+        'points' => $level->getPoints(),
+        'bonus' => $level->getBonus(),
+        'bonus_dec' => $level->getBonusDec(),
+        'bonus_fix' => $level->getBonusFix(),
+        'flag' => $level->getFlag(),
+        'hint' => $level->getHint(),
+        'penalty' => $level->getPenalty()
+  		);
+  		array_push($all_levels_data, $one_level);
+  	}
+  	return $all_levels_data;
+  }
+
+  public async function genCategories(): Awaitable<mixed> {
+  	$all_categories_data = array();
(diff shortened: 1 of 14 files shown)


src/controllers/exporter/ExporterController.php
```diff
@@ -0,0 +1,99 @@
+<?hh // strict
+
+abstract class ExporterController {
+  abstract public function genData(string $type): Awaitable<mixed>;
+
+  public async function genTeams(): Awaitable<mixed> {
+		$all_teams_data = array();
```
gsingh93: You're mixing tabs/spaces in this file, convert all the tabs to spaces.

src/controllers/exporter/ExporterController.php
```diff
@@ -0,0 +1,99 @@
...
+  	$all_teams = \HH\Asio\join(Team::genAllTeams());
```
gsingh93: Why not just use `await`?

src/controllers/exporter/JSONExporterController.php
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  public static function genJSON(mixed $data): string {	
+    return json_encode($data, JSON_PRETTY_PRINT);   
+  }
+
+  public static function sendJSON(mixed $data, string $json_file='fbctf.json'): void {
+    header('Content-Type: application/json;charset=utf-8');
+    header('Content-Disposition: attachment; filename='.$json_file);
+    print self::genJSON($data);
```
gsingh93: Let's stick with `echo` to be consistent.

src/controllers/importer/ImporterController.php
```diff
@@ -0,0 +1,134 @@
+<?hh // strict
+
+abstract class ImporterController {
+  abstract public function importData(string $input_filename, string $type): ?array<string, mixed>;
+
+  public async function processTeams(array<string, mixed> $teams): Awaitable<bool> {
```
gsingh93: Maybe the argument type should be `array<string, array<string, mixed>>`? That also might fix your hack errors. Same argument for the other functions.

src/controllers/importer/ImporterController.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+    return true;
+  }
+
+  public async function importLogos(string $input_filename): Awaitable<bool> {
+    $logos = $this->importData($input_filename, 'logos');
+    return await $this->processLogos($logos);
+  }
+
+  public async function importGame(string $input_filename): Awaitable<bool> {
```
gsingh93: One problem I see with the way you're doing this function is that it's possible for part of the import to succeed, which might be bad. i.e. we could successfully process logos, and then fail to import teams.

src/controllers/importer/JSONImporterController.php
```diff
@@ -0,0 +1,31 @@
+<?hh // strict
+
+class JSONImporterController extends ImporterController {
+	<<__Override>>
```
gsingh93: Convert the tabs to spaces in this file.

src/models/Control.php
```diff
@@ -112,22 +111,93 @@ class Control extends Model {
...
+    $game = await $exporter->genData('game');
+    JSONExporterController::sendJSON($game, 'fbctf_game.json');
+    exit();
+  }
+
+  public static async function exportTeams(): Awaitable<void> {
+    $exporter = new JSONExporterController();
+    $teams = await $exporter->genData('teams');
+    JSONExporterController::sendJSON($teams, 'fbctf_teams.json');
+    exit();
```
gsingh93: Can you explain why we need these exit calls? We should just structure the code so they're not needed.

src/models/Token.php
```diff
@@ -57,6 +57,7 @@ private static function generate(): string {
 
   // Create token.
   public static async function genCreate(): Awaitable<void> {
+    error_log('TOKEN!');
```
gsingh93: remove

src/static/js/admin.js
```diff
@@ -381,14 +381,113 @@ function databaseBackup() {
...
+  $('#import-levels_file').change(function() {
+    submitImport('levels_file', 'import_levels');
+  });
+}
+
+// Export and download current teams
+function exportCurrentTeams() {
+  var csrf_token = $('input[name=csrf_token]')[0].value;
+  var action = 'export_teams';
+  var url = 'index.php?p=admin&ajax=true&action=' + action + '&csrf_token=' + csrf_token;
```
gsingh93: I don't like the idea of putting the csrf token in the URL. Can we make this a POST parameter?

gsingh93:
This doesn't need to be done in this PR, but I think there's a cleaner way to do the data model to exported data conversion. We can have `Exportable`/`Importable` interfaces that classes like `Team` implements. Then when we want to make a new piece of data exportable/importable, we only have to implement that interface for that new piece of data.

facebook-github-bot-4:
Thank you for your pull request.  As you may know, we require contributors to sign our Contributor License Agreement, and we don't seem to have you on file and listed as active anymore.  In order for us to review and merge your code, please email cla@fb.com with your details so we can update your status.

facebook-github-bot-8:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

Issue #153, opened 2016-06-28
Increase default MySQL connection limit


gsingh93:
By default, MySQL has an extremely low connection limit. This causes problems with running the CTF with around 100 people, even though the platform can support many more. We should include a `my.cnf` file that sets `max_connections` to a larger number (maybe 1000?), as stated [here](https://dev.mysql.com/doc/refman/5.5/en/too-many-connections.html).

javuto:
I think this is totally reasonable. Given that the default value is 151 (see it with the command ```show variables like "max_connections";```) we can easily and safely put it to 500 at least.

javuto:
We should definitely deploy our own ```my.cnf``` and also enable binary logs.

JamieH:
I used FBCTF for a CTF with around 50 people and had a whole bunch of issues with max_connections. Another solution would be to use persistent connections or connection pooling perhaps?

gsingh93:
We've fixed various performance issues, and we're continuing to do more. After thinking about this issue, I don't think we should make any changes.

max_connections is definitely very low by default, but we shouldn't be overwriting the entire my.cnf file when we provision. The user could have already put a number of tweaks in there and we don't want to override those. Because of that, I'm going to close this.

PR #179, opened 2016-08-04
#153 Increase MYSQL Connection Limit

Erikt360:
Line 19 Changed


gsingh93:
So turns out the fix for this was actually the my.cnf file discussed in a comment on #153. I don't think the solution in the original comment was correct, so I'm going to close this. Thanks for the PR anyways!

PR #187, opened 2016-08-14
DropKick version bump to fix XSS

javuto:
There was a problem with DropKick and Cross Site Scripting, due to the element being assigned the text instead of the encoded innerHTML. See https://github.com/Robdel12/DropKick/pull/338 for the fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   },
   "dependencies": {
     "d3": "^3.5.16",
-    "dropkickjs": "^2.1.8",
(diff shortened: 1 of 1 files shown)


facebook-github-bot-5:
Thank you for your pull request.  As you may know, we require contributors to sign our Contributor License Agreement, and we don't seem to have you on file and listed as active anymore.  In order for us to review and merge your code, please email cla@fb.com with your details so we can update your status.

PR #197, opened 2016-08-28
Use a unix socket instead of TCP for fastcgi

gsingh93:
This gives a bit of performance boost.


javuto:
Merge away!

wisco24:
I am now having issues  with this change.  geting a nginx 502 error when installing on AWS small instance.

PR #208, opened 2016-09-02
New PR to add import/export after rebase

javuto:
Continuing the original pull request for import/export here (https://github.com/facebook/fbctf/pull/150) after leaving it for too long and having way too many conflicts.


src/controllers/ajax/AdminAjaxController.php
```diff
@@ -410,7 +425,55 @@ protected function getActions(): array<string> {
...
+        await Control::exportLevels();
+        return Utils::ok_response('Success', 'admin');
+      case 'export_categories':
+        await Control::exportCategories();
+        return Utils::ok_response('Success', 'admin');
+      case 'import_game':
+        $result = await Control::importGame();
+        if ($result) {
+          return Utils::ok_response('Success', 'admin');
+        } else {
```
iangcarroll: It's usually cleaner to just remove the `else` if the conditional is going to return.

```
if ($result) {
    return value;
}

return value2;
```

Alternatively you could make these a one-liner with ternary operators, though I don't think it'd be cleaner.
```
return $result ? Utils::ok_response('Success', 'admin') : Utils::error_response('Error importing', 'admin');
```

src/controllers/exporter/ExporterController.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+    return $all_teams_data;
+  }
+
+  public async function genLevels(): Awaitable<mixed> {
+    $all_levels_data = array();
+    $all_levels = \HH\Asio\join(Level::genAllLevels());
+
+    foreach ($all_levels as $level) {
+      $entity = \HH\Asio\join(Country::gen($level->getEntityId()));
```
iangcarroll: Might be better to await these at the same time.

src/controllers/exporter/ExporterController.php
```diff
@@ -0,0 +1,99 @@
...
+        'used' => $logo->getUsed(),
+        'enabled' => $logo->getEnabled(),
+        'protected' => $logo->getProtected()
+      );
+      array_push($all_logos_data, $one_logo);
+    }
+    return $all_logos_data;
+  }
+
+  public async function genAll(): Awaitable<mixed> {
```
iangcarroll: This will always return an array, right? `Awaitable<array>` would be better.

src/controllers/exporter/ExporterController.php
```diff
@@ -0,0 +1,99 @@
...
+    $levels = $this->genLevels();
+    $teams = $this->genTeams();
+    $logos = $this->genLogos();
+    $all_data = array(
+      'categories' => $categories,
+      'levels' => $levels,
+      'teams' => $teams,
+      'logos' => $logos
+    );
+    return $all_data;
```
iangcarroll: You can just return the array itself:
```php
return array(
  'categories' => $categories,
  'levels' => $levels,
  'teams' => $teams,
  'logos' => $logos
);
```

src/models/Control.php
```diff
@@ -106,21 +105,93 @@ class Control extends Model {
...
   }
 
-  public static function importGame(): void {}
+  public static async function importGame(): Awaitable<bool> {
+    $files = Utils::getFILES();
+    if ($files->contains('game_file')) {
+      $importer = new JSONImporterController();
+      $result = await $importer->importGame($files['game_file']['tmp_name']);
+      return $result;
+    } else {
```
iangcarroll: Same thing as above; you can just drop the `else`s.

src/controllers/exporter/ExporterController.php
javuto: I don't understand what you mean here

src/controllers/exporter/ExporterController.php
javuto: It will have to be ```array<mixed>``` then, otherwise you get ```You cannot have an array without generics in strict mode (Typing[4045])```

src/controllers/exporter/ExporterController.php
CaptureTheBox: One of the main benefits of await is you can do things like `await gena(array(gen1, gen2))`, and gen1 and gen2 will be awaited in parallel. He's saying `gen` all the entity IDs in one command. In any case, no new code should ever use `join`.

src/controllers/exporter/ExporterController.php
CaptureTheBox: array<mixed> is fine too.

gsingh93:
This PR includes some commits that are already in master. Can you fix that?

Are you planning on implementing the `Importable` interface we discussed in the last PR?

iangcarroll:
Composer is warning about class name conflicts; this could be a problem (`importers` vs `importer`, `exporters` vs `exporter`):

```
Warning: Ambiguous class resolution, "JSONImporterController" was found in both "/var/www/fbctf/src/controllers/importers/JSONImporterController.php" and "/var/www/fbctf/src/controllers/importer/JSONImporterController.php", the first will be used.
Warning: Ambiguous class resolution, "JSONExporterController" was found in both "/var/www/fbctf/src/controllers/exporters/JSONExporterController.php" and "/var/www/fbctf/src/controllers/exporter/JSONExporterController.php", the first will be used.
```

javuto:
@iangcarroll both ```importer``` and ```exporter``` should have been removed, only ```importers``` and ```exporters``` are valid

freeqaz:
This is awesome!

Issue #209, opened 2016-09-05
Latest update broke Docker(?)


frite:
I believe that the latest commit (32a9f47d488895925cf187e80340ac2ff143cacc) broke hhvm for docker.
```docker -v
Docker version 1.12.0, build 8eab29e```

```
$ docker exec -it e24668d642e3 bash
root@e24668d642e3:~# service nginx status
 * nginx is running
root@e24668d642e3:~# service mysql status
 * /usr/bin/mysqladmin  Ver 8.42 Distrib 5.5.50, for debian-linux-gnu on x86_64
Copyright (c) 2000, 2016, Oracle and/or its affiliates. All rights reserved.

Oracle is a registered trademark of Oracle Corporation and/or its
affiliates. Other names may be trademarks of their respective
owners.

Server version 		5.5.50-0ubuntu0.14.04.1
Protocol version       	10
Connection     		Localhost via UNIX socket
UNIX socket    		/var/run/mysqld/mysqld.sock
Uptime:			29 sec

Threads: 1  Questions: 110  Slow queries: 0  Opens: 71  Flush tables: 1  Open tables: 61  Queries per second avg: 3.793
root@e24668d642e3:~# service hhvm status
 * hhvm is not running
root@e24668d642e3:~#
```

Vagrant on the other hand works fine.


I see there's a build job here https://travis-ci.org/facebook/fbctf/builds . Maybe it can be expanded to run tests for docker as well.  

Regards

qazbnm456:
@frite I've noticed that it's the permission problem of hhvm's socket, which cause the broken image. I've sent PR already, and you might want to follow the commit to fix this problem out. :)

frite:
Awesome, thanks a million @qazbnm456

qazbnm456:
@frite Not at all.

PR #210, opened 2016-09-07
fix https://github.com/facebook/fbctf/issues/209

qazbnm456:
fix https://github.com/facebook/fbctf/issues/209

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ RUN apt-get update \
   ca-certificates \
   && chown www-data:www-data $HOME \
   && ./extra/provision.sh -m $MODE -c $TYPE -k $KEY -C $CRT -D $DOMAIN -e $EMAIL -s `pwd` --docker \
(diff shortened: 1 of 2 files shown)


javuto:
Awesome! Great work @qazbnm456

gsingh93:
Thanks for the fix. Does the hhvm process run as root at the moment? We might want to consider running it as www-data for better security, and so we don't have to chmod in the future. In any case, this is fine for now.

qazbnm456:
It runs as www-data, but socket is owned by root. Therefore, I have to `chown` socket to www-data to make it writable by nginx. 

I found this issue has been reported here (https://github.com/facebook/hhvm/issues/6336).

aarondanielson:
@gsingh93 Can we take another look at this?  @qazbnm456 has found the solution, perhaps adding a command-line argument to specify the socket owner is a good way to deal with this issue.

PR #212, opened 2016-09-08
Hungarian characters added to font files

markszabo:
The Hungarian characters ő and ű were missing from the font files, so I added them:
selection_012
(Before on the left, after on the right)

I'm new to font editing, so I will share my entire process:

1. Opened the fonts in [FontForge](https://fontforge.github.io/en-US/)
2. Copied the existing ó,ú,Ó,Ú characters to ő,ű,Ő,Ű
3. Edited the characters (duplicated and aligned the accents)
4. Saved it in `sfd` format, since FontForge only supports that
5. Converted from `sfd` to `woff` using [this](https://everythingfonts.com/sfd-to-woff) online tool
6. Generated the `woff2` files from the `woff` using an [other tool](https://everythingfonts.com/woff-to-woff2) from the same website

I have tested it under Ubuntu with the newest Firefox and Chrome, and it works.


javuto:
Merged! Thanks for the PR @markszabo

PR #216, opened 2016-09-12
Mitigation for broken HHVM 3.15 package

javuto:
Latest HHVM package (3.15) for Ubuntu 14.04/16.04 is broken so this mitigates the impact by installing the last good package (3.14.5).
Issue in HHVM already created: https://github.com/facebook/hhvm/issues/7333

### DIFF
--- a/extra/lib.sh
+++ b/extra/lib.sh
@@ -226,6 +226,13 @@ function install_hhvm() {
   log "Installing HHVM"
   sudo apt-get update
   package hhvm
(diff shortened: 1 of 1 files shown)


eeeehaw:
This did not resolve my same problem.  I have HHVM 3.14.5 installed with FBCTF on both Azure and Google Cloud VMs running Ubuntu 14.04LTS.  Initial provision runs fine, but after restarting the VM I then get 502 Bad Gateway from nginx, and 'sudo service hhvm status' says that hhvm is not running after the restart.  See Issue https://github.com/facebook/fbctf/issues/234 .

Issue #165, opened 2016-07-08
Leader board graph not showing line during competition


acjcool:
Hi,

I've been playing some quiz games and noticed that the leader board graph isn't updating. The team scores are but not the graph. Is there something that I should switch on in the config screen or something? It used to work before I pulled the latest from github
![fbctf](https://cloud.githubusercontent.com/assets/3783663/16681412/dd45dd42-44f4-11e6-8bd5-7a2b82a62ee7.PNG)

no1rscyber:
I noticed the same fault on my last game. I thought it might of been that I started the game without any teams logged in? Were your teams all logged in and ready to go when you started the game?

acjcool:
Hi, sorry for the late reply. Both teams were actively playing. The table updates and the start->end blocks were filling up. The graph did nothing so getting worried as I'm getting asked with every demo why the graph isn't working. Any ideas or is someone else experiencing something similar?

jeszicawii:
Hi! 
We are having trouble with the base scores updating, however haven't tested the scores. 
It seems like maybe something should be polling (updating the graph, bases score etc) that maybe isn't?

Do you think its likely these issues could be related?

javuto:
Hello! Thanks for your patience and for submitting the issues. I think I have identified the problem and I will commit a fix soon. The progressive scoreboard not being updated it is related with not having active teams at the moment of hitting the "Begin Game" button. @jeszicawii I don't think the bases issue is related but I am investigating that one. I'll update that one in the other issue.

rdenman:
Another issue I noticed with the scoreboard is that once I've close it, I'm no longer able to interact with the map (i.e. cannot zoom in/out, drag to a new position, or click on countries).

Issue #180, opened 2016-08-09
Progressive scoreboard to run without active teams


javuto:
The progressive scoreboard is expected to be kicked off (after hit Begin Game) with all the teams active. If the Game starts with no teams active, due to the registration still open, the data does not flow to the ```progressive_log``` table. Investigate. This may be related with the issue with bases not working properly #169

Issue #169, opened 2016-07-18
Can't get bases running.


jeszicawii:
Hi all,

We have had bases running in the past however can't seem to set them up anymore, just wondering if we're missing anything silly, a sanity check.
We've followed the instructions on the wiki and even when we visit the port in browser we get the team name, however the map isn't updating.

This is using the vagrant file provided in the git repo via git clone.

javuto:
Hey @jeszicawii, I am currently investigating this one and I'll keep you posted. Thanks for your patience :)

jeszicawii:
@javuto Thank you!

jeszicawii:
Any update on this?

gsingh93:
Sorry for the delay, but we probably won't have any updates until after defcon.

alolis:
Do bases work on latest version?

PR #259, opened 2016-10-04
Fix progressive scoreboard and bases

irqbank:
When a new game is started, `src/models/Progressive.php` starts `scripts/progressive.php` in the background by passing it as a parameter to hhvm from the shell. The file `scripts/progressive.php` includes `autoload.php`, which initiates a chain of inclusions, most of which use `$_SERVER['DOCUMENT_ROOT']` to determine the file path. Since the `$_SERVER` variable is not available when hhvm is called from the shell, the shell command errors out, and the progressive scoreboard doesn't work. The workaround is to manually include the required php files for `scripts/progressive.php` to work.

It seems to me that this is related to #165. Can anyone please reject my analysis by showing that the progressive scoreboard works for them with the latest code?

New open issue: the background process is not killed when the game is stopped.

### DIFF
--- a/src/scripts/bases.php
+++ b/src/scripts/bases.php
@@ -1,6 +1,12 @@
 <?hh
 
-require_once ('/var/www/fbctf/vendor/autoload.php');
+require_once('/var/www/fbctf/src/Db.php');
+require_once('/var/www/fbctf/src/Utils.php');
+require_once('/var/www/fbctf/src/models/Model.php');
+require_once('/var/www/fbctf/src/models/Importable.php');
+require_once('/var/www/fbctf/src/models/Exportable.php');
+require_once('/var/www/fbctf/src/models/Level.php');
+require_once('/var/www/fbctf/src/models/Configuration.php');
 
 $conf_game = \HH\Asio\join(Configuration::gen('game'));
 while ($conf_game->getValue() === '1') {
--- a/src/scripts/progressive.php
+++ b/src/scripts/progressive.php
@@ -1,6 +1,10 @@
 <?hh
 
-require_once ('/var/www/fbctf/vendor/autoload.php');
+require_once('/var/www/fbctf/src/Db.php');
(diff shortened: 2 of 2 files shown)


javuto:
This could be a good approach to fix not just the progressive scoreboard, but what lives in the ```src/scripts``` folder. Let me do some testing and verify the fix works but one thing to improve would be to not hardcode ```/var/www/fbctf``` because that is set when running ```provision.sh```.

irqbank:
@javuto Good point, I hadn't actually checked the other files in the `src/scripts` folder. Since this folder currently contains `bases.php`, this implies that bases aren't working either. Correcting the includes for `src/scripts/bases.php` likely addresses #169 and #180. I will try to submit a PR for this if I can find the time.

irqbank:
Committed a similar fix for bases.php. Unfortunately I did not have the time to improve even further by removing the hardcoded paths.

javuto:
Please, do rebase since I merged a change for ```bases.php``` and ```progressive.php```

inche-ali:
please check PR #310

#hacktoberfest :)

noobiexco:
I am having an issue with the progressive scoreboard and am unsure how to fix. I have tried to change the bases.php and progressive.php files to include the required files, however this seems to break base functionality and the base no longer submits the json to the server for scoring. Currently I have the base sending the request and the server accepting and saying that the team specified in the SCORE_POINTS captures the base. The scoreboard shows the correct score on the line, but not in the numbers underneath. Creating a new user or a flag capture seems to force the leaderboard to update and then correct score is shown. I have seen the issue with the language.php file and autoload.php. Should I manually hardcode the autoload files in there similar to what was done to bases and progressive in this thread? 
Really dont want to set up a script to create and delete a user every minute for the sake of forcing a refresh
Thanks for any help.

justinwray:
Should be closable with with the merge of #379.

PR #264, opened 2016-10-05
Disallow running scripts from web

fredemmott:
Runs fine from terminal, but:

```
fredemmott-mbp1:extra fredemmott$ curl -kI https://localhost/scripts/bases.php
HTTP/1.1 405 Method Not Allowed
Server: nginx
Date: Wed, 05 Oct 2016 16:49:58 GMT
Content-Type: text/html
Connection: keep-alive
X-Powered-By: HHVM/3.14.5
Vary: Accept-Encoding
```

### DIFF
--- a/src/scripts/bases.php
+++ b/src/scripts/bases.php
@@ -1,5 +1,10 @@
 <?hh
 
+if (php_sapi_name() !== 'cli') {
+  http_response_code(405); // method not allowed
+  exit(0);
+}
+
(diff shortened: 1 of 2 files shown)


javuto:
Thanks for the PR @fredemmott 
This could have been bad for CTFs that are based on bases, since the scoring might happen when the script runs, so a team could use this in its own benefit.

PR #267, opened 2016-10-07
Use the correct CSS selector to find out if names are required

### LINKED ISSUE #86: Player Names
I've enabled player names, which allows for input during registration, but once submitted I can't find where to check who was on each team.

In the admin panel under Teams it says "No Team Names" when looking at the 'Names' tab above the team. Also, on the gameboard, when clicking on a team logo in the 'Teams' window nothing is listed under 'team_members'.

Is this a bug?

Hopefully the above makes sense...

koddsson:
The selector `#register_names` doesn't seem to be in the DOM.

This change replaces that check with one that checks for the value of the hidden input named `action`. If that value is `register_names` we make sure to collect the names and add them to the AJAX response to the server.

Fixes #86

### DIFF
--- a/src/static/js/fb-ctf.js
+++ b/src/static/js/fb-ctf.js
@@ -2289,7 +2289,7 @@ function setupInputListeners() {
     $body = $('body');
 
     $('#login_button').click(Index.loginTeam);
-    var names_required = $('#register_names').length > 0;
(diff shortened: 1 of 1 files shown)


javuto:
Thanks for the PR! I verified the fix and it makes the registration to collect the names and emails.

PR #271, opened 2016-10-11
Don't pretty print DataController JSON data

Orvid:
Because it can (and does) quite readily more than double the size of the data being sent, and is completely unnecessary.


javuto:
Makes sense. Thank you for the PR! 👍

Issue #251, opened 2016-10-02
Translate fbctf platform to Basque


javuto:
Add support for the platform in Basque:

https://github.com/facebook/fbctf/tree/master/src/language

joxeankoret:
Some friends and I are probably taken the translation. With a bit of luck, it will be done by next week.

joxeankoret:
...or by today. @mieltxo and I from [EuskalHack](http://euskalhack.org/) just finished the translation: 
[lang_eu-beta1.php.txt](https://github.com/facebook/fbctf/files/505953/lang_eu-beta1.php.txt)

We skipped on purpose only one part because IANAL:

```
By participating in the contest, you agree to release Facebook and its employees, and the hosting organization from any and all liability, claims or actions of any kind whatsoever for injuries, damages or losses to persons and property which may be sustained in connection with the contest. You acknowledge and agree that Facebook et al is not responsible for technical, hardware or software failures, or other errors or problems which may occur in connection with the contest.
```

If you still want us to translate this too, please consider that we aren't lawyers so there might be some inconsistencies.

kartik95:
Added Basque Translation. PR #272

kartik95:
Will my PR be merged anytime soon? Or does it need any changes? @javuto

iangcarroll:
They may need some time to review it (and the other PRs) before it's merged, depending on how busy they are. 🙂

gsingh93:
Sorry for the delay, we've had issues with translations in the past, so we're finding the right people to review them.

PR #272, opened 2016-10-12
Added Basque translation.

kartik95:
Referencing #251 

Added Basque translation of fbctf.


joxeankoret:
OK, I have signed it. Thanks for adding the translation!

PR #274, opened 2016-10-13
Update provision.sh

prakhar987:


javuto:
If you have a question, please open an issue instead of sending a Pull Request with code that does not add any value at all. Thanks.

Issue #241, opened 2016-10-01
Filter module reset view on refresh, maintaining the applied filter


javuto:
When the filter module reloads (top right corner), it does reset the filter in the view but the filter applied remains. For example, if we do apply a filter to display only the "Quiz" challenges, it will refresh and go back to "All", but the view is not consistent.

gsingh93:
I think making a new cookie makes sense to persist this over page refreshes.

slayerjain:
expertise in Php is necessary ?

javuto:
Yes, this requires previous experience with PHP/Hack.

PR #278, opened 2016-10-17
Persistent filters on refresh, added new cookies

javuto:
Implementation for #241, this pull request adds a new javascript module, ```filter.js```, that persist the values of the filters module in a set of cookies.

### DIFF
--- a/src/static/js/fb-ctf.js
+++ b/src/static/js/fb-ctf.js
@@ -23,7 +23,7 @@ var widgetsList = [
 ];
 Widget.rememberWidgets(widgetsList);
 
-// This will be populated in loadSavedFilterModule
+// This Object will be populated in loadSavedFilterModule
 var filterList = {};
 
 // Capture enter key presses on login and registration forms
--- a/src/static/js/fb-ctf.js
+++ b/src/static/js/fb-ctf.js
@@ -2,6 +2,7 @@
 
 var Index = require('./index');
 var Widget = require('./widget');
+var Filter = require('./filter');
 var Utils = require('./utils');
 var Modal = require('./modal');
 var Slider = require('./slider');
@@ -20,8 +21,10 @@ var widgetsList = [
   'Filter',
   'Game Clock'
 ];
-
 Widget.rememberWidgets(widgetsList);
+
+// This will be populated in loadSavedFilterModule
+var filterList = {};
 
 // Capture enter key presses on login and registration forms
 $(document).on('keypress', 'input', function(e) {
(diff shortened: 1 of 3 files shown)


src/static/js/fb-ctf.js
```diff
@@ -20,9 +21,11 @@ var widgetsList = [
   'Filter',
   'Game Clock'
 ];
-
 Widget.rememberWidgets(widgetsList);
 
+// This will be populated in loadSavedFilterModule
+var filterList = {};
```
gsingh93: Since this is a list, let's initialize with square brackets for clarity.

src/static/js/filter.js
```diff
@@ -0,0 +1,115 @@
+var $ = require('jquery');
+
+function isFilterSet(filterName) {
+  return document.cookie.search(filterName) >= 0;
+}
+
+module.exports = {
+  getFilterName: function(filterSelector, filters) {
+    for (var key in filters) {
```
gsingh93: Don't use `for..in` with JavaScript arrays: https://stackoverflow.com/questions/500504/why-is-using-for-in-with-array-iteration-a-bad-idea

src/static/js/filter.js
```diff
@@ -0,0 +1,115 @@
...
+    var expires = "expires=" + d.toUTCString();
+    document.cookie = filterName + "=" + filterValue + "; " + expires;
+  },
+
+  getFilterState: function(filterName) {
+    var name = filterName + "=";
+    var ca = document.cookie.split(';');
+    for (var i = 0; i < ca.length; i++) {
+      var c = ca[i];
+      while (c.charAt(0) == ' ') c = c.substring(1);
```
gsingh93: Always use curly braces

src/static/js/filter.js
```diff
@@ -0,0 +1,115 @@
...
+      if (c.indexOf(name) === 0) return c.substring(name.length, c.length);
+    }
+    return '';
+  },
+
+  rememberFilters: function(filters) {
+    for (var key in filters) {
```
gsingh93: Don't use `for..in`

src/static/js/filter.js
```diff
@@ -0,0 +1,115 @@
...
+      if (isFilterSet(key)) {
+        var ftype;
```
gsingh93: Here and everywhere else for new code, let's use `let` instead of `var`.

src/static/js/filter.js
javuto: I could use ```Object.keys``` and then iterate through those...

Issue #185, opened 2016-08-13
Improve session management


gsingh93:
We currently do DB writes very often because of the session `write` function, and we do DELETE queries as well because of `destroy` and `gc`. We should reduce the number of times these are called.

javuto:
Maybe we should also look at the values of [gc_probability](http://php.net/manual/en/session.configuration.php#ini.session.gc-probability), [gc_divisor](http://php.net/manual/en/session.configuration.php#ini.session.gc_divisor) and [gc_maxlifetime](http://php.net/manual/en/session.configuration.php#ini.session.gc_maxlifetime). They will impact on how many times ```gc``` will be called.

PR #279, opened 2016-10-17
Major Reduction of MySQL Queries

justinwray:
* Creation of the MultiTeam class, extending the Team class and overloading certain class methods to combine database queries and retain data retrieved from the database for subsequent usage.  Overall, this class dramatically reduces the queries to the teams table.  Specifically, this ensures queries are only called once per page request.

* MultiTeam::genTeam() replaces the Team instantiation with a method that utilizes the teams database results stored within the Multi::Team class.  All Team::genTeam() calls have been updated to utilize MultiTeam::genTeam().  This reduces the number of queries when building multiple team objects from 1(n) to 1.

* MultiTeam::genLeaderboard() replaces the method to retrieve a sorted array of Team objects used for scoring displays.  Previously this would query the database 1(n) times per team. It is now performing one query per page request that includes any leaderboard display.

* MultiTeam::genPointsByType() replaces the retrieval of scores by various type (flag, quiz, base).  This is primarily used by the Scoreboard display.  Previously, when building the scoring display, the database would be queried 3(n) times per team. The new method only queries the database once per page request.

* MultiTeam::genAllActiveTeams() replaces the method used to find all active teams.  While this was only performing a single query previously, the data was not retained.  In the past, if there were multiple calls to Team::genAllActiveTeams() within the same code execution (request) it would perform multiple queries.  The new overloaded method now reduces this query against the database to once per page request.

* MultiTeam::genAllVisibleTeams() replaces the method used to find all visible teams.  While this was only performing a single query previously, the data was not retained.  In the past, if there were multiple calls to Team:: genAllVisibleTeams() within the same code execution (request) it would perform multiple queries.  The new overloaded method now reduces this query against the database to once per page request.

* MultiTeam::genWhoUses replaces the team-by-logo retrieval.  This method will return the teams who use a particular logo, which is used on both the Scoreboard and the administrative team page.  In the past, this would perform one query per logo.  The queries have been reduced from hundreds to one, per page request.

* MultiTeam::genCompletedLevel() replaces the method used to find particular levels that have been completed, and who completed the levels.  Previously this would query the database 1(n) times, per level, and is now performing 1 query per page request that includes any leaderboard display

* MultiTeam::genFirstCapture() replaces the method used to determine who captured a particular scoring level first.  Previously this was called for all levels 1(n) times and is now called once per page request.

* Updated relevant calls to the aforementioned Team class methods which have now been overloaded by the MultiTeam class.  The team class has been left intact as the possibility exists that there will be valid use cases where specific, singular team, queries are valid.  It is also possible that specific, singular team functionality or extension of the functionality may be needed in the future, along with backward compatibility. Alternatively, the MultiTeam class could be merged into Team in the future.

* Added Router::getRequestedPage(), Router::isRequestAjax(), and Router::isRequestRouter() to the Router class.

* Router::getRequestedPage() will return either the "page" $_REQUEST variable, if set, or the "p" $_REQUEST variable.  This can be used to determine which page was requested.  This provides filter functionality based off of the requested page or utilization of the value for logging.

* Router::isRequestAjax() will return a boolean.  This method can be used whenever specific functionality needs to be filtered based on whether the request was an XHR or not.  Note, that currently, not all XHR go through the Router class.

* Router::isRequestRouter() will return a boolean.  This method can be used whenever specific functionality needs to be filtered based on whether or not the request went through the Router or not.

* SessionUtils::read() has been updated to utilize a single method to return the session data.  Previously there was first a request to the session table within the database to determine if the session existed, then a second query would retrieve the session data.  Now the checking of the session and return of the session data take place within a single database query.  The new method is Session:genSessionDataIfExists().

* Session::getLastPageAccess() has been added to the Session class.  This method will return the last page that was requested. The last page requested is retrieved from Router::getRequestedPage() and is therefore based on the "p" or "page" value from $_REQUEST.

* The sessions table within the database has been updated to include a "last_page_access" column, which now contains the last page that was accessed when the session was last updated.

* Both Session::genCreate() and Session::genUpdate() have been updated to include the new database column on insert or update.

* Session::genUpdate() and Session::getCleanup() are no longer executing on every request.  These functions now filter out any request which is an AJAX request or doesn't go through the Router class.  Currently, modals are still being tracked, but this can be changed if further query reduction is desired.  Overall, this results in a massive reduction in queries to the session table, which was happening on a continual basis due to the scoreboard updates via XHR.  This should resolve facebook/fbctf#185.

* The admin Sessions page has been updated to display the last page access.  This was added to make up for the loss of fidelity on the last access time, as XHR will no longer update the session and therefore no longer update the last access time.  However, using a combination of the last access time and the last access page, it can be determined what a user is doing, including if they are idling on the Scoreboard display.  This seems to be a favorable compromise, especially given the massive query reduction it provides.

* Updated relevant language files for the new Last Page Access display.

### DIFF
--- a/database/test_schema.sql
+++ b/database/test_schema.sql
@@ -160,6 +160,7 @@ CREATE TABLE `sessions` (
   `team_id` int(11) NOT NULL,
   `created_ts` timestamp NOT NULL DEFAULT 0,
   `last_access_ts` timestamp NOT NULL,
+   `last_page_access` text NOT NULL,
   PRIMARY KEY (`id`)
 ) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
--- a/database/test_schema.sql
+++ b/database/test_schema.sql
@@ -160,7 +160,7 @@ CREATE TABLE `sessions` (
   `team_id` int(11) NOT NULL,
   `created_ts` timestamp NOT NULL DEFAULT 0,
   `last_access_ts` timestamp NOT NULL,
-   `last_page_access` text NOT NULL,
+  `last_page_access` text NOT NULL,
   PRIMARY KEY (`id`)
 ) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
--- a/src/Router.php
+++ b/src/Router.php
@@ -100,7 +100,6 @@ class Router {
 
   // Check to see if the request is going through the router
   public static function isRequestRouter(): bool {
-    if (self::getRequestedPage() == "index") return false;
-    else return true;
+    return self::getRequestedPage() != "index";
   }
 }
--- a/src/Router.php
+++ b/src/Router.php
@@ -84,4 +84,23 @@ class Router {
         throw new NotFoundRedirectException();
     }
   }
+
+  public static function getRequestedPage(): string {
+    $page = idx(Utils::getGET(), 'page') ?: idx(Utils::getGET(), 'p');
+    if (!is_string($page)) {
+      $page = 'index';
+    }
+
+    return strval($page);
+  }
+
+  public static function isRequestAjax(): bool {
+    return Utils::getGET()->get('ajax') === 'true';
+  }
+
+  // Check to see if the request is going through the router
+  public static function isRequestRouter(): bool {
+    if (self::getRequestedPage() == "index") return false;
+    else return true;
+  }
 }
--- a/src/models/MultiTeam.php
+++ b/src/models/MultiTeam.php
@@ -1,0 +1,211 @@
+<?hh // strict
+
+class MultiTeam extends Team {
+
+  private static Map<int, Team> $all_teams = Map {};
+  private static array<Team> $team_leaderboard = [];
+  private static Map<int, Map<string, int>> $team_points_by_type = Map {};
+  private static array<Team> $all_active_teams = [];
+  private static array<Team> $all_visible_teams = [];
+  private static Map<string, array<Team>> $teams_by_logo = Map {};
+  private static Map<int, array<Team>> $teams_by_completed_level = Map {};
+  private static Map<int, Team> $first_team_captured_by_level = Map {};
+  
+  private static function setAllTeams(Map<int, Team> $all_teams): void {
+  	self::$all_teams = $all_teams;
+  }
+  
+  private static function setLeaderboard(array<Team> $team_leaderboard): void {
+    self::$team_leaderboard = $team_leaderboard;
+  }
+
+  private static function setPointsByType(Map<int, Map<string, int>> $team_points_by_type): void {
+    self::$team_points_by_type = $team_points_by_type;
+  }
+
+  private static function setAllActiveTeams(array<Team> $all_active_teams): void {
+  	self::$all_active_teams = $all_active_teams;
+  }
+  
+  private static function setAllVisibleTeams(array<Team> $all_visible_teams): void {
+    self::$all_visible_teams = $all_visible_teams;
+  }
+  
+  private static function setAllTeamsByLogo(Map<string, array<Team>> $teams_by_logo): void {
+  	self::$teams_by_logo = $teams_by_logo;
+  }
+  
+  private static function setTeamsByCompletedLevel(Map<int, array<Team>> $teams_by_completed_level): void {
+    self::$teams_by_completed_level = $teams_by_completed_level;
+  }
+  
+  private static function setFirstTeamCapturedByLevel(Map<int, Team> $first_team_captured_by_level): void {
+  	self::$first_team_captured_by_level = $first_team_captured_by_level;
+  }
+  
+  private static async function genTeamArrayFromDB(
+    string $query,
+  ): Awaitable<Vector<Map<string, string>>> {
+  	$db = await self::genDb();
+  	$result = await $db->query($query);
+  	
+  	return $result->mapRows();
+  }
+  
+  // All teams.
+  public static async function genAllTeamsCache(
+    bool $refresh = false,
+    ): Awaitable<Map<int, Team>> {
+      if ((count(self::$all_teams) === 0) || ($refresh)) {
+        $all_teams = Map {};
+        $teams = await self::genTeamArrayFromDB('SELECT * FROM teams');
+        foreach ($teams->items() as $team) {
+          $all_teams->add(Pair {intval($team->get("id")), Team::teamFromRow($team)});
+        }
+        self::setAllTeams($all_teams);
+       }
+    return self::$all_teams;
+  }
+  
+  public static async function genTeam(
+    int $team_id,
+    bool $refresh = false,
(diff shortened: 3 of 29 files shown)


database/test_schema.sql
```diff
@@ -160,6 +160,7 @@ CREATE TABLE `sessions` (
   `team_id` int(11) NOT NULL,
   `created_ts` timestamp NOT NULL DEFAULT 0,
   `last_access_ts` timestamp NOT NULL,
+   `last_page_access` text NOT NULL,
```
gsingh93: nit: looks like there's an extra space at the start of this.

src/Router.php
```diff
@@ -84,4 +84,23 @@ class Router {
...
+    return strval($page);
+  }
+
+  public static function isRequestAjax(): bool {
+    return Utils::getGET()->get('ajax') === 'true';
+  }
+
+  // Check to see if the request is going through the router
+  public static function isRequestRouter(): bool {
+    if (self::getRequestedPage() == "index") return false;
```
gsingh93: Yea, the code would be better if it was written as @iangcarroll suggested. But I'm not sure why we're saying all requests that don't have the page `index` go through the router.

src/models/MultiTeam.php
```diff
@@ -0,0 +1,211 @@
...
+            $points_by_type->add(Pair {intval($team->get("id")), $type_pair});
+          }
+        }
+      }
+      /* HH_IGNORE_ERROR[4110] */
+      self::setPointsByType(new Map($points_by_type));
+    }
+    /* HH_IGNORE_ERROR[4064] */
+    if ((array_key_exists(intval($team_id), self::$team_points_by_type)) && (self::$team_points_by_type->contains($team_id)) && (self::$team_points_by_type->get($team_id)->contains($type))) return intval(self::$team_points_by_type->get($team_id)->get($type));
+    else return intval(0);
```
gsingh93: Always use curly braces for if statements.

src/models/MultiTeam.php
```diff
@@ -0,0 +1,211 @@
...
+        self::setAllTeams($all_teams);
+       }
+    return self::$all_teams;
+  }
+  
+  public static async function genTeam(
+    int $team_id,
+    bool $refresh = false,
+  ): Awaitable<Team> {
+  	await self::genAllTeamsCache($refresh);
```
gsingh93: Fix indentation

src/models/MultiTeam.php
```diff
@@ -0,0 +1,211 @@
...
+    /* HH_IGNORE_ERROR[4110] */
+    return self::$all_teams->get($team_id);
+  }
+  
+  // Leaderboard order.
+  public static async function genLeaderboard(
+    bool $refresh = false,
+  ): Awaitable<array<Team>> {
+
```
gsingh93: remove extra newline

src/models/MultiTeam.php
```diff
@@ -0,0 +1,211 @@
...
+    /* HH_IGNORE_ERROR[4110] */
+    return self::$all_teams->get($team_id);
+  }
+  
+  // Leaderboard order.
+  public static async function genLeaderboard(
+    bool $refresh = false,
+  ): Awaitable<array<Team>> {
+  	
+  	if ((count(self::$team_leaderboard) === 0) || ($refresh)) {
```
gsingh93: Fix indentation

src/models/MultiTeam.php
```diff
@@ -0,0 +1,211 @@
...
+      $all_teams = await self::genAllTeamsCache();    
+      
+      $teams_by_logo = array();
+      foreach ($all_teams as $team) {
+        $teams_by_logo[$team->getLogo()][] = $team;
+      }
+      self::setAllTeamsByLogo(new Map($teams_by_logo));
+    }
+    /* HH_IGNORE_ERROR[4110] */
+    if ((count(self::$teams_by_logo) !== 0) && (array_key_exists($logo, self::$teams_by_logo))) return self::$teams_by_logo->get("$logo");
```
gsingh93: You don't need quotes around $logo

justinwray:
@gsingh93 Thank you for the feedback!

I will reply to the relevant commits inline.  I just wanted to respond to the overall comments here.

We skipped Memcache in this PR to simplify the review.  We had planned to introduce Memcache in this functionality during an upcoming effort.  With that said, if you would prefer the Memcache implementation to be a part of this PR that's not a problem at all.

As it relates to the expiration of the cached data within the MultiTeam class:  Given the current implementation, the database queries will be performed once per relevant request.  So the data is going to be up-to-date as of the time of the browser's request, and the initial invocation of the method.  You are right that we are never setting $refresh, but we included it as there are use cases (that we are not currently overloading or replacing) where you would need to get a new copy of the data from the database.  For example, you would want to set $refresh when you are dealing with a request which is both updating and retrieving the team data.  In that case, you would want to call the methods with $refresh after the update function completes.

We can certainly get a test created for MultiTeam.

Lastly, apologies for the combined PR.  It was our original intention to submit two PRs here.  However, we ended up concurrently working through both and wanted to get the changes submitted for feedback.  In the future we will separate PRs and if you would prefer we can separate this PR as well.

database/test_schema.sql
justinwray: No problem, we will remove this.  On a side note, we originally ran the format.sh script, but it changed a large number of upstream files which appear to be non-complaint.  Would you like a PR for re-formatting this code and existing code?

src/Router.php
justinwray: We will update this.  I'm not sure why the shorthand versions were used in some cases and the longer form in others.  We will go back over the code and see if there are any other similar cases.

As for the Router::isRequestRouter() method, we're using the Router::getRequestedPage() method to retrieve the "p" or "page" request variables.  For simplicity, if neither 'p' nor 'page' are set, we are returning a default of "index."  None of the non-router requests have either the "p" or "page" variables set, so we are then determining the request did not go through the Router.

However, this would be inaccurate on an index.php request.  Although, in the case that it is a request directly to the root index.php, it does not impact the current usage of the methods.  We could add a check for the requested file to determine if it is a true index request.  I think such a change would be better and clarify the functionality further.

src/models/MultiTeam.php
justinwray: No problem.  We will update this across the board.

src/models/MultiTeam.php
justinwray: Understood.  We will clean that up.

justinwray:
Commit aa730c2 has been added to the PR based on the changes requested by @gsingh93.

javuto:
Code looks good to me, just a couple of things:
- Translation to Portuguese was just merged, so you will need to check that one for the new string.
- I don't fully understand the usage of ```isRequestRouter```, maybe the function name is just confusing.

justinwray:
Merged with the upstream project and added the Portuguese translation for the new display string (Last Page Access).

This last commit should cover all outstanding changes requested by @gsingh93 and @javuto and be compatible with the current upstream project.

@javuto The isRequestRouter() method is used to determine if a page has gone through the Router class or not.  Some requests go directly to a PHP file, while others go through the Router class first before being routed to the requested code.  This method is currently being used to determine if the Session should be updated or not.

gsingh93:
Thanks guys!

PR #285, opened 2016-10-20
Forbid correct answer submissions on inactive levels

dannybd:
This part of the code still allows correct answer submissions on inactive levels, which can lead to weird UI states. This commit prevents that behavior. I also made `$level_id` and `$answer` into variables to improve readability of this part.


javuto:
Just reproduced the bug and tested the fix. All good. Thank you for the PR!

dannybd:
You're very welcome. Thanks for merging!

PR #288, opened 2016-10-26
Fix index error in genFirstCapture in Teams.php

bitcoinctf:
genFirstCapture would throw an error due to the query return no rows, but the return statement requires one row (due to the [0] ). This would happen when a disabled or invisible user was had the first (but not only) capture. I used a second sub-query to fix this as the rest of this code base appears to favor sub-queries instead of joins.

### DIFF
--- a/src/models/Team.php
+++ b/src/models/Team.php
@@ -561,7 +561,7 @@ class Team extends Model implements Importable, Exportable {
     $db = await self::genDb();
     $result =
       await $db->queryf(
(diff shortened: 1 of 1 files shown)


javuto:
This is a great find and fix. Thanks for the Pull Request! As soon as the CLA is signed, I will merge it.

javuto:
Ping @bitcoinctf

gsingh93:
Thanks!

Issue #245, opened 2016-10-02
Translate fbctf platform to Russian


javuto:
Add support for the platform in Russian.

light2yellow:
@javuto I can try my best.

nehagup:
i would like to solve this issue!

light2yellow:
@javuto see #249

PR #298, opened 2016-10-27
Update lang_basque.php

vishnubhagwan:
#245


iangcarroll:
I don't even know if this is valid. Still pointless.

javuto:
Pointless and just spam.

Issue #155, opened 2016-06-29
Don't Set MySQL Root Password


Ttech:
When doing the configuration for production, the setup script sets the MYSQL Password and the 'ctf' user password. 

The 'ctf' user by default has the password 'ctf', as well, which might be insecure.

Ideally, these credentials should be generated randomly and given to the setup to be used.

gsingh93:
I don't think this is a huge issue because the MySQL only listens on localhost by default. I do think it's a good idea for users to change it on their own, but it will definitely create more confusion for people if we do it automatically in the provision script (scroll through the closed issues and see how many are about what the password is).

Ttech:
I agree - but thats why I was suggesting a output message of "here are the passwords" don't forget! 
This turned out to be quite a problem when setting this up for external listening on the mysql server. 
Since I had to stop everything and start it up to get the password of root set properly and then while I was doing that changed the ctf password.

gsingh93:
So we actually randomly generate a password for the admin user when you provision in prod mode, and that gets printed out at the end of provisioning. A lot of the issues we had after launch were because people didn't read the output at the end and missed the password.

Maybe we can have a flag for the provision script (off by default) that randomly generates the password?

javuto:
We could define the mysql password as a parameter in the provision script or have it generated. For now is just hardcoded and this does not really go well with previous installations of MySQL.

rdenman:
For those that do want to change the default MySQL password (which should be anyone actually running a competition), you'll need to change the /var/www/fbctf/settings.ini DB_PASSWORD parameter. If you do not edit settings.ini, then when you change the password in MySQL you're platform will no longer load/work.

PR #305, opened 2016-10-28
Don't Set MySQL Root Password (#155)

shlykov:
Added generation random password of root database user and ctf apps user for production mode
Added output notify message  with root database password

### DIFF
--- a/extra/provision.sh
+++ b/extra/provision.sh
@@ -39,9 +39,12 @@
 # We want the provision script to fail as soon as there are any errors
 set -e
 
-DB="fbctf"
-U="ctf"
-P="ctf"
+# Database credentials.
+DB_NAME="fbctf"
+DB_USER="ctf"
+
+#Default passwords, used in development mode only
+DB_PASWWORD="ctf"
 P_ROOT="root"
 
 # Default values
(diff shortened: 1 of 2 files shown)


gsingh93:
I was actually a bit hesitant about this at first, considering the number of people who did not see the admin password printed out during provision, the fact that the DB password isn't as important as the admin password, because most production deployments of fbctf don't expose the DB to the outside world, and finally because forgetting the admin password isn't a big deal because it's easy to reset, but reseting the mysql password is a pain in the ass.

But since we write the password to the settings.ini file, this should be fine. Going to wait for @javuto to give his opinion first.

javuto:
This is a good idea and I agree with @gsingh93 that could be a bit risky. Instead of doing it by default, we could have the option as a command line switch. Also, this PR can be extended to take existing username/password for the MySQL access.

extra/provision.sh
```diff
@@ -39,9 +39,12 @@
...
 
-DB="fbctf"
-U="ctf"
-P="ctf"
+# Database credentials.
+DB_NAME="fbctf"
+DB_USER="ctf"
+
+#Default passwords, used in development mode only
+DB_PASWWORD="ctf"
```
javuto: ```DB_PASSWORD```

PR #309, opened 2016-10-29
Updated Memcached Implementation

justinwray:
- Moved setMCRecords(), getMCRecords(), and invalidateMCRecords() to the Model class.  These methods will be utilized in any Model class in which Memcached is implemented.
- MC_KEY, MC_EXPIRE, and MC_KEYS have been converted to protected static properties on the Model class, which can then be overloaded by the children classes.  These static parent properties allow each class to define the relevant Memcached settings needed.
- MC_EXPIRE now defaults to 0 (indefinite), this can be overridden by a child class if the need arises.
- The Memcached for MultiTeam and ScoreLog are invalidated on demand.  This invalidation takes place when teams are modified, created, or deleted and when scoring events are triggered.
- Fixed an issue where the full class cache wasn't being invalidated if no parameter was provided to invalidateMCRecords().
- Updated ScoreLog::genPreviousScore() to reduce queries and utilize Memcached.  Previously the method was calling 2(n) queries per level.  Now the query has been consolidated to a single query, and the data is stored in Memcached and only runs after invalidation on a scoring event.

### DIFF
--- a/src/models/Model.php
+++ b/src/models/Model.php
@@ -41,7 +41,6 @@ abstract class Model {
   protected static function getMCRecords(string $key): mixed {
     $mc = self::getMc();
     $mc_result = $mc->get(static::$MC_KEY.static::$MC_KEYS->get($key));
-    /* HH_IGNORE_ERROR[4110] */
     return $mc_result;
   }
 
--- a/src/models/Model.php
+++ b/src/models/Model.php
@@ -3,6 +3,10 @@
 abstract class Model {
   protected static Db $db = MUST_MODIFY;
   protected static Memcached $mc = MUST_MODIFY;
+  protected static string $MC_KEY = 'MUST_MODIFY';
+  protected static int $MC_EXPIRE = 0; //Defaults to indefinite cache life
+
+  protected static Map<string, string> $MC_KEYS = Map {};
 
   protected static async function genDb(): Awaitable<AsyncMysqlConnection> {
     if (self::$db === MUST_MODIFY) {
@@ -23,5 +27,32 @@ abstract class Model {
       self::$mc->addServer($host, $port);
     }
     return self::$mc;
+  }
+
+  protected static function setMCRecords(string $key, mixed $records): void {
+    $mc = self::getMc();
+    $mc->set(
+      static::$MC_KEY.static::$MC_KEYS->get($key),
+      $records,
+      static::$MC_EXPIRE,
+    );
+  }
+
+  protected static function getMCRecords(string $key): mixed {
+    $mc = self::getMc();
+    $mc_result = $mc->get(static::$MC_KEY.static::$MC_KEYS->get($key));
+    /* HH_IGNORE_ERROR[4110] */
+    return $mc_result;
+  }
+
+  public static function invalidateMCRecords(?string $key = null): void {
+    $mc = self::getMc();
+    if (is_null($key)) {
+      foreach (static::$MC_KEYS as $key_name => $mc_key) {
+        $mc->delete(static::$MC_KEY.static::$MC_KEYS->get($key_name));
+      }
+    } else {
+      $mc->delete(static::$MC_KEY.static::$MC_KEYS->get($key));
+    }
   }
 }
--- a/src/models/Model.php
+++ b/src/models/Model.php
@@ -3,8 +3,8 @@
 abstract class Model {
   protected static Db $db = MUST_MODIFY;
   protected static Memcached $mc = MUST_MODIFY;
-  protected static string $MC_KEY = 'MUST_MODIFY';
-  protected static int $MC_EXPIRE = 0; //Defaults to indefinite cache life
+  protected static string $MC_KEY = MUST_MODIFY;
+  protected static int $MC_EXPIRE = 0; // Defaults to indefinite cache life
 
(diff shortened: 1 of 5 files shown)


src/models/Model.php
```diff
@@ -3,6 +3,10 @@
 abstract class Model {
   protected static Db $db = MUST_MODIFY;
   protected static Memcached $mc = MUST_MODIFY;
+  protected static string $MC_KEY = 'MUST_MODIFY';
```
gsingh93: You don't need to make this a string, `MUST_MODIFY`, is a special constant.

src/models/Model.php
```diff
@@ -24,4 +28,31 @@ protected static function getMc(): Memcached {
...
+  protected static function getMCRecords(string $key): mixed {
+    $mc = self::getMc();
+    $mc_result = $mc->get(static::$MC_KEY.static::$MC_KEYS->get($key));
+    /* HH_IGNORE_ERROR[4110] */
+    return $mc_result;
+  }
+
+  public static function invalidateMCRecords(?string $key = null): void {
+    $mc = self::getMc();
+    if (is_null($key)) {
```
gsingh93: nit: while `is_null` is equivalent to `!== null`, we've been using the latter so far, so let's stick with it.

src/models/ScoreLog.php
```diff
@@ -69,28 +75,49 @@ private static function scorelogFromRow(Map<string, string> $row): ScoreLog {
...
-        );
-    } else {
-      $result =
-        await $db->queryf(
-          'SELECT COUNT(*) FROM scores_log WHERE level_id = %d AND team_id = %d',
-          $level_id,
-          $team_id,
-        );
+    $mc_result = self::getMCRecords('LEVEL_CAPTURES');
+    if ((!$mc_result) || (count($mc_result) === 0) || ($refresh)) {
```
gsingh93: I can understand the parens around the middle conditional (even if it isn't necessary), but I'm not a fan of the parens around the first and last conditional.

src/models/Model.php
```diff
@@ -24,4 +28,31 @@ protected static function getMc(): Memcached {
...
+      static::$MC_KEY.static::$MC_KEYS->get($key),
+      $records,
+      static::$MC_EXPIRE,
+    );
+  }
+
+  protected static function getMCRecords(string $key): mixed {
+    $mc = self::getMc();
+    $mc_result = $mc->get(static::$MC_KEY.static::$MC_KEYS->get($key));
+    /* HH_IGNORE_ERROR[4110] */
```
gsingh93: We probably haven't been doing this enough, but if we're going to use an `HH_IGNORE`, let's specify why we're ignoring it in the comment, i.e. `/* HH_IGNORE_ERROR[4110]: xyz */`

src/models/ScoreLog.php
```diff
@@ -69,28 +75,49 @@ private static function scorelogFromRow(Map<string, string> $row): ScoreLog {
...
+      $level_captures = Map {};
+      $result = await $db->queryf('SELECT level_id, team_id FROM scores_log');
+      foreach ($result->mapRows() as $row) {
+        if ($level_captures->contains(intval($row->get("level_id")))) {
+          $level_capture_teams =
+            $level_captures->get(intval($row->get("level_id")));
+          /* HH_IGNORE_ERROR[4064] */
```
gsingh93: Same comment about `HH_IGNORE_ERROR` above. I think it'll help readability and prevent bugs.

src/models/ScoreLog.php
```diff
@@ -69,28 +75,49 @@ private static function scorelogFromRow(Map<string, string> $row): ScoreLog {
...
+        }
+      }
+      self::setMCRecords('LEVEL_CAPTURES', new Map($level_captures));
     }
-
-    if ($result->numRows() > 0) {
-      invariant($result->numRows() === 1, 'Expected exactly one result');
-      return intval($result->mapRows()[0]['COUNT(*)']) > 0;
+    $level_captures = self::getMCRecords('LEVEL_CAPTURES');
+    /* HH_IGNORE_ERROR[4062] */
```
gsingh93: You don't have to specify a comment for each of these `HH_IGNORE_ERROR`s if they're all the same. Just do one. Hack has an `/* UNSAFE_BLOCK */`, but that's caused me more problems than it's solved.

justinwray:
@gsingh93 Updates based on your feedback have been committed.

src/models/ScoreLog.php
gsingh93: Looks like you missed this one. Other than that everything looks good.

Also, tests will be broken until #315 is merged, so we'll merge this after that's merged.

justinwray: Oops.  Sorry about that.  It's the same suppression as found on line 109 (below this one).

Also, work on fixing tests seems to have been carried out in parallel, as #315 seems to be a duplicate of #309.  #315 can likely be closed if #309 fully resolves the test issues.

justinwray: Edited the previous comment to clarify:  #323 seems to be a duplicate of #315.

gsingh93:
Can you rebase now that the tests are fixed?

justinwray:
@gsingh93 Rebase complete!

PR #311, opened 2016-10-30
Fix formatting of error page glitching.

gregprosser:
The current XHP results in the literal string {tr('ERROR')} ending up in the output.  XHP doesn't quote attributes with "".


gregprosser:
Admittedly I didn't spin up a VM to test this, but this matches other parts of the source (and this is the only place we do ="{ in the code, seemingly.

gregprosser:
Repro: navigate to /index.php?page=error on your instance and notice the "glitching" has extra letters.  This will fix itself and just be "glitching" the same text afterwards.

gsingh93:
Thanks :)

Issue #312, opened 2016-10-30
multiple mycli repo added in /etc/apt/sources.list


inche-ali:
this issue occurs during re-install
# hacktoberfest :)

javuto:
No longer a problem, since mycli is now installed from pip

PR #313, opened 2016-10-30
Avoid multiple entry in /etc/apt/sources.list when re-install

inche-ali:
fix for issue #312

### DIFF
--- a/extra/lib.sh
+++ b/extra/lib.sh
@@ -51,13 +51,6 @@ function repo_osquery() {
   sudo add-apt-repository "deb [arch=amd64] https://osquery-packages.s3.amazonaws.com/trusty trusty main"
 }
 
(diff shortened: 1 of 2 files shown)


gsingh93:
It looks like MyCLI now supports installing from pip. I think it would make more sense to install `pip` and then install `mycli` from there.

javuto:
Agree with @gsingh93, better install pip, which should be just ```package python-pip```, then install it there.

inche-ali:
modified the provision.sh & lib.sh files...now use pip to install mycli
done rebase to fbctf/master too..

javuto:
Looks good to me, cleaner and no need to add a new repository. @gsingh93?

gsingh93:
Merged, thanks!

PR #315, opened 2016-10-30
Test and import game fix

inche-ali:
Fix for test script related with PR #232  - updated some testcase, test_schema, test data
Fix for game import game - missing input element in AdminController 
                                         - add checking using "instanceof KeyedContainer"

### DIFF
--- a/src/models/Control.php
+++ b/src/models/Control.php
@@ -120,23 +120,35 @@ class Control extends Model {
 
   public static async function importGame(): Awaitable<bool> {
     $data_game = JSONImporterController::readJSON('game_file');
-    if ($data_game instanceof KeyedContainer) {
+    if ($data_game) {
       $logos = array_pop(must_have_idx($data_game, 'logos'));
+      if (!$logos) {
+        return false;
+      }
       $logos_result = await Logo::importAll($logos);
       if (!$logos_result) {
         return false;
       }
       $teams = array_pop(must_have_idx($data_game, 'teams'));
+      if (!$teams) {
+        return false;
+      }
       $teams_result = await Team::importAll($teams);
       if (!$teams_result) {
         return false;
       }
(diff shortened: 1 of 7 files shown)


src/models/Control.php
```diff
@@ -119,24 +119,24 @@ class Control extends Model {
   }
 
   public static async function importGame(): Awaitable<bool> {
-    $data_game = self::readJSON('game_file');
-    if ($data_game) {
-      $logos = must_have_idx($data_game, 'logos');
+    $data_game = JSONImporterController::readJSON('game_file');
+    if ($data_game instanceof KeyedContainer) {
```
gsingh93: If the data isn't a KeyedContainer we should return `false` early. You can also `error_log` what the error was so we can find it in the logs later.

src/models/Control.php
```diff
@@ -119,24 +119,24 @@ class Control extends Model {
...
+      $logos = array_pop(must_have_idx($data_game, 'logos'));
       $logos_result = await Logo::importAll($logos);
```
gsingh93: Does this pass typechecking (`hh`)? The documentation says `array_pop` can return null, but `Logo::importAll` can't take null.

inche-ali:
Thanks for reviewing my fix. I will try to modify it

inche-ali:
OK..I fixed it.
please review

#hacktoberfest :)

gsingh93:
Looks good to me. Just to confirm, you did run `hh_client` on it this time, right? Also, run `extra/format.sh` in case there are any formatting changes that need to be made (make sure to rebase first).

inche-ali:
ahh.. thanks for reminding me, forgot that to run that format.sh & hh_client.. 
I've pushed another changes, after fix some type checking.. should be okay now..

#hacktoberfest :)

inche-ali:
done rebasing..

gsingh93:
Thanks!

PR #325, opened 2016-11-01
Combined Asynchronous Awaitables

justinwray:
* Combined multiple sequential awaits based on the documentation provided here:  https://docs.hhvm.com/hack/async/awaitables#awaiting__batching-awaitables

* The combined configuration requests provide minor performance improvements, due to the existing efficient usage of Memcached.

* The combined Scoring Logs requests provides a clear performance improvement, especially when there is a large amount of scoring activity within the logs.

### DIFF
--- a/src/controllers/AdminController.php
+++ b/src/controllers/AdminController.php
@@ -3336,19 +3336,26 @@ class AdminController extends Controller {
 
         if ($results->contains('team') && $results->contains('level')) {
           $team = $results->get('team');
+          invariant($team !== null, 'Team should not be null');
+          invariant($team instanceof Team, '\$team should be of type Team');
+
           $level = $results->get('level');
-          /* HH_IGNORE_ERROR[4064]: Object can never be null due to line 3321 */
+          invariant($level !== null, 'Level should not be null');
+          invariant(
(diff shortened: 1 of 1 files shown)


justinwray:
@gsingh93 Rebased this one as well!

src/controllers/AdminController.php
```diff
@@ -3255,40 +3326,55 @@ class={$highlighted_color}
...
+          'team' => MultiTeam::genTeam($gamelog->getTeamId()),
+          'level' => Level::gen($gamelog->getLevelId()),
+        };
+
+        $results = await \HH\Asio\m($awaitables);
+
+        if ($results->contains('team') && $results->contains('level')) {
+          $team = $results->get('team');
+          $level = $results->get('level');
+          /* HH_IGNORE_ERROR[4064]: Object can never be null due to line 3321 */
```
gsingh93: Instead of using HH_IGNORE_ERROR here (and any other place where this applies), it's better to convince Hack that something can't be null. You can do something like `invariant($level !== null, "Level should not be null")`.

If you ever want to assert something is of the right type, you can use `is_array` for arrays or `instanceof` for objects.

Issue #213, opened 2016-09-08
Making Multiple Announcements Creates Clutter


rdenman:
Every time a new announcement is made, it simply stacks right on top of the old ones. It would be nice for there to be a bit more space between them, or possibly a date above each announcement to separate them. As is, it can get a bit cluttered. It may also be nice if there could be some sort of notification (could be as simple as a number next to announcements in header) when a new announcement is made so players can see them in a timely manner.

javuto:
This is definitely an issue. I can add a bit of space between announcements and also the time when the announcement was made. For the notification, I can put some flashing in the Announcements module, if it is open. If the module is closed/collapsed, I can make it to open and flash too. What do you think @gsingh93 ?

gsingh93:
Sounds good. Having the date/time as a separator would actually be a good feature.

PR #330, opened 2016-11-03
Announcements have now timestamp and are less clutered

javuto:
Announcements did not have a proper format and they ended up as clutter and not really helpful. Issue is #213. Now they have the timestamp and separation.


javuto:
14962924_10154682983519730_407558640_n

Issue #329, opened 2016-11-03
Docker build fails due to missing C compiler


freeqaz:
Howdy y'all,

I've been following this project for a while and I just updated to the latest master. It looks like something is breaking now when I run the Docker install.

Here's my system information.
```shell
> docker -v
Docker version 1.12.2-rc1, build 45bed2c, experimental

> uname -av
Darwin my-host 15.6.0 Darwin Kernel Version 15.6.0: Thu Sep  1 15:01:16 PDT 2016; root:xnu-3248.60.11~2/RELEASE_X86_64 x86_64

> git rev-parse HEAD
2ee8f3a334e1aa9349a0b2afff0827130f7bd72a
```

The output from when I run `docker build --build-arg MODE=dev -t="fbctf_in_dev" .` is [here](https://gist.github.com/freeqaz/ee5d65932820c8009a56383daeff77a7).

Thank you!

zjuchenyuan:
modify Dockerfile can overcome this problem
I changed Dockerfile "RUN" line and finally passed build:
(adding build-essential curl python-dev)

RUN apt-get update \
  && apt-get install -y \
  build-essential python-dev curl nginx git language-pack-en python-pip htop memcached git npm nodejs-legacy \
  rsync \
  ca-certificates \
  && chown www-data:www-data $HOME \
  && ./extra/provision.sh -m $MODE -c $TYPE -k $KEY -C $CRT -D $DOMAIN -e $EMAIL -s `pwd` --docker \
  && rm -f /var/run/hhvm/sock \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*

PR #332, opened 2016-11-04
Install correct dependencies in dev

freeqaz:
Fixes issue #329 by installing dependencies.

Which version of python do we need, 2 or 3? This works with both, but I want to only include what we need.

Reviewers: @javuto


javuto:
Thanks for the PR! I think we can stick to python 2 so just ```python-dev``` should be enough.

javuto:
Can you sign the CLA so we can merge this? Thanks!

freeqaz:
Signed!

PR #339, opened 2016-11-05
Added Memcached Implementation to Level and Control

justinwray:
* Updated Memcached Implementation on Country, ScoreLog, and Team.

* Updated Level::genWhoUses(), Level::genCheckStatus(), Level::genCheckBase() to reduce queries and utilize Memcached.  Previously each method was calling one query per country.  Now the methods have been updated to consolidate the various queries into a single query for all levels.  The resulting data is stored in Memcached and queries only runs after cache invalidation.

* Updated Level::genAllLevels(), Level::genAllActiveLevels(), and Level::genAllActiveBases() to utilize Memcached.  These methods were previously only calling a single query each, but they are used often on the Gameboard updates.

* Updated Level::genAllTypeLevels() to utilize Memcached.  This method was previously performing a single query per level type, but the method is used frequently on the Gameboard updates.

* Updated Level::gen() to utilize Memcached.  This method is used to generate a single level, and was called 1(n) times per level.  Now all levels are stored in Memcached and retrieved as needed.

* The Memcached data for Level is invalidated on demand.  This invalidation takes place when Levels are updated, added, or removed.

* Updated the Memcached implementation within the Country class to bring it inline with the other Memcached code.

* Updated the Control class to implement Memcached on Control::genAllActivity().  This method is utilized to generate the Activity log used on the Gameboard.  The method previously executed a single query frequently on Gameboard updates.  Now the query is executed once, with the results stored in Memcached.  The cache is invalidated on any Scoring event or modification to the relevant dependencies, such as changes to Levels, Countries, Teams, etc.

* Updated the ScoreLog class to invalidate Control and MultiTeam caches when scores are reset or a scoring event is triggered.

* Updated ScoreLog::genPreviousScore() to move the database connection inside of the condition block in which it is utilized.  This eliminates the database connection when no query will take place.

* Updated the Team class to invalidate the Activity log cache when a team is updated, deleted, or the visibility is changed.

* Removed caching from Country::genAll() as the relevant caches have been moved to the appropriate methods within the class.

* The Memcached data for Country is invalidated on demand.  This invalidation takes place when countries are updated, such as a change in status or usage.

### DIFF
--- a/src/controllers/AdminController.php
+++ b/src/controllers/AdminController.php
@@ -3341,10 +3341,7 @@ class AdminController extends Controller {
 
           $level = $results->get('level');
           invariant($level !== null, 'Level should not be null');
-          invariant(
-            $level instanceof Level,
-            'level should be of type Level',
-          );
+          invariant($level instanceof Level, 'level should be of type Level');
 
           $country = await Country::gen($level->getEntityId());
 
--- a/src/models/Control.php
+++ b/src/models/Control.php
@@ -5,7 +5,7 @@ class Control extends Model {
   protected static string $MC_KEY = 'control:';
 
   protected static Map<string, string>
-    $MC_KEYS = Map {"ALL_ACTIVITY" => "activity"};
+    $MC_KEYS = Map {'ALL_ACTIVITY' => 'activity'};
 
   public static async function genStartScriptLog(
     int $pid,
@@ -263,8 +263,12 @@ class Control extends Model {
         );
       self::setMCRecords('ALL_ACTIVITY', $result->mapRows());
     }
-    /* HH_IGNORE_ERROR[4110]: getMCRecords returns a 'mixed' type, HHVM is unsure of the type at this point */
-    return self::getMCRecords('ALL_ACTIVITY');
+    $all_activity = self::getMCRecords('ALL_ACTIVITY');
+    invariant(
+      $all_activity instanceof Vector,
+      'all_activity should be of type Vector',
+    );
+    return $all_activity;
   }
 
   public static async function genResetBases(): Awaitable<void> {
--- a/src/models/Control.php
+++ b/src/models/Control.php
@@ -262,13 +262,13 @@ class Control extends Model {
           'SELECT scores_log.ts AS time, teams.name AS team, countries.iso_code AS country, scores_log.team_id AS team_id FROM scores_log, levels, teams, countries WHERE scores_log.level_id = levels.id AND levels.entity_id = countries.id AND scores_log.team_id = teams.id AND teams.visible = 1 ORDER BY time DESC LIMIT 50',
         );
       self::setMCRecords('ALL_ACTIVITY', $result->mapRows());
+      return $result->mapRows();
     }
-    $all_activity = self::getMCRecords('ALL_ACTIVITY');
     invariant(
-      $all_activity instanceof Vector,
-      'all_activity should be of type Vector',
+      $mc_result instanceof Vector,
+      'cache return should be of type Vector',
     );
-    return $all_activity;
+    return $mc_result;
   }
 
   public static async function genResetBases(): Awaitable<void> {
--- a/src/models/Control.php
+++ b/src/models/Control.php
@@ -1,6 +1,12 @@
 <?hh // strict
 
 class Control extends Model {
+
+  protected static string $MC_KEY = 'control:';
+
+  protected static Map<string, string>
+    $MC_KEYS = Map {"ALL_ACTIVITY" => "activity"};
+
   public static async function genStartScriptLog(
     int $pid,
     string $name,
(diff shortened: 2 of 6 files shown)


src/models/Control.php
```diff
@@ -1,6 +1,12 @@
 <?hh // strict
 
 class Control extends Model {
+
+  protected static string $MC_KEY = 'control:';
+
+  protected static Map<string, string>
+    $MC_KEYS = Map {"ALL_ACTIVITY" => "activity"};
```
gsingh93: This is a nitpick, but here and wherever you use double quotes for strings, convert them to single quotes. We only use double quotes when we explicitly want to have variables evaluated in strings.

src/models/Country.php
```diff
@@ -87,31 +92,28 @@ public function getTransform(): string {
...
 
   private static async function genAll(
     string $sql,
-    string $mcKey,
   ): Awaitable<array<Country>> {
-    $mc = self::getMc();
-    $mc_result = $mc->get($mcKey);
+    $db = await self::genDb();
+    $all_countries = Map {};
+    /* HH_IGNORE_ERROR[4110] */
```
gsingh93: What's this HH_IGNORE_ERROR for?

src/models/Country.php
```diff
@@ -123,91 +125,107 @@ function($a, $b) {
...
-    $db = await self::genDb();
-
-    $result = await $db->queryf(
-      'SELECT COUNT(*) FROM levels WHERE entity_id = %d AND active = 1',
-      $country_id,
-    );
-
-    invariant($result->numRows() === 1, 'Expected exactly one result');
-    return intval(firstx($result->mapRows())['COUNT(*)']) > 0;
+    return Level::genWhoUses($country_id) != null;
```
javuto: I believe this is an error? ```genIsActiveLevel``` has not much to do with ```genWhoUses```

justinwray: Level::genWhoUses() returns a Level based on the input of the country Id, or _null_ otherwise.  Country::genIsActiveLevel(), by supplying a country Id, is used to determine if a country is in use.  So in this case, if Level::genWhoUses($country_id) returns _null_ the country is not in use, whereas, if the method returns a Level (_not null_) the country is in use.

src/models/Country.php
justinwray: This HH_IGNORE_ERROR was already in the code, from commit d5846c5. However, it appears to be related to the _$sql_ parameter being passed to _$db->queryf()_.  The _queryf()_ method is expecting a type ofHH\FormatString, but is being provided a standard string here.

With that said, none of the calls to Country::genAll() are providing any conditions or inputs into the queries.  Thus there is no need for a formatted string, and _$db->query()_ can be used instead of _$db->queryf()_.

This will be updated in the next commit and will remove the need for both of the HH_IGNORE_ERROR usages in this case.

src/models/Country.php
```diff
@@ -123,91 +126,134 @@ function($a, $b) {
...
   }
 
   // Get a country by id.
-  public static async function gen(int $country_id): Awaitable<Country> {
-    $db = await self::genDb();
-
-    $result = await $db->queryf(
-      'SELECT * FROM countries WHERE id = %d LIMIT 1',
-      $country_id,
+  /* HH_IGNORE_ERROR[4110]: Claims - It is incompatible with void because this async function implicitly returns Awaitable<void>, yet this returns Awaitable<Country> and the type is checked on line 230 */
```
gsingh93: The reason it's saying this, is because if the final `if` statement is false, the function will end without returning a value.

src/controllers/AdminController.php
```diff
@@ -3341,10 +3341,7 @@ class={$highlighted_color}
 
           $level = $results->get('level');
           invariant($level !== null, 'Level should not be null');
-          invariant(
-            $level instanceof Level,
-            'level should be of type Level',
-          );
+          invariant($level instanceof Level, 'level should be of type Level');
```
justinwray: This is a minor change.  Not sure why the `format.sh` script is treating this differently now.

justinwray:
This PR has been updated with the discussed changes and the race condition fix from #352.  This specific PR (#339) is a prerequisite for #340.  You can test a full merge of all of the Memcached related PRs via [WraySec/fbctf/tree/full_merge](/WraySec/fbctf/tree/full_merge).

src/models/Level.php
```diff
@@ -647,18 +814,43 @@ private static function levelFromRow(Map<string, string> $row): Level {
...
+        invariant($level instanceof Level, 'level should be of type Level');
+        return $level;
+      }
+    } else {
+      invariant(
+        $mc_result instanceof Map,
+        'cache return should be of type Map',
+      );
+      invariant($mc_result->contains($level_id) !== false, 'level not found');
+      if ($mc_result->contains($level_id)) {
```
gsingh93: Since we have the `invariant`, let's remove the if statement and the HH_IGNOREs at the top. Same with the other places where we do this.

src/models/Level.php
```diff
@@ -180,38 +207,86 @@ private static function levelFromRow(Map<string, string> $row): Level {
...
     } else {
-      return false;
+      invariant(
+        $mc_result instanceof Map,
+        'cache return should be of type Map',
+      );
+      if ($mc_result->contains($level_id)) {
+        $level = $mc_result->get($level_id);
+        invariant($level instanceof Level, 'level should be type of Level');
+        if ($level->type == 'base') {
```
gsingh93: Use triple equals.

justinwray: Not a problem!

src/models/Level.php
justinwray: Not a problem!

justinwray:

src/models/Country.php
```diff
@@ -123,91 +120,155 @@ function($a, $b) {
...
-    $result = await $db->queryf(
-      'SELECT * FROM countries WHERE id = %d LIMIT 1',
-      $country_id,
-    );
-
-    invariant($result->numRows() === 1, 'Expected exactly one result');
-    $new_country =
-      await self::countryFromRow(firstx($result->mapRows())->toArray());
-    return $new_country;
+  /* HH_IGNORE_ERROR[4110]: Lines #246 and #260 prevent this function from failing to return */
```
gsingh93: Do you actually need this or did you just forget to fix this function like the others?

justinwray:
@gsingh93 The last HH_IGNORE_ERROR has been removed in favor of the invariant check.

PR #340, opened 2016-11-05
Added Memcached Implementation to HintLog

justinwray:
* Updated HintLog::genPreviousHint() to utilize Memcached.  Previously the method was calling 2(n) queries per level.  Now the query has been consolidated to a single query, and the data is stored in Memcached.  The single query will now only run after invalidation due to a hint being used.

* Updated the GameAjaxController class to invalidate the caches for scoring (the MultiTeam class) on hint usage.

* Updated GameAjaxController class to remove the MultiTeam and ScoreLog cache invalidations on a valid scoring event, these are now handled internally within the relevant classes.  This change is dependent on PR: facebook/fbctf#339

### DIFF
--- a/src/controllers/ajax/GameAjaxController.php
+++ b/src/controllers/ajax/GameAjaxController.php
@@ -60,13 +60,6 @@ class GameAjaxController extends AjaxController {
             );
             // Update teams last score
             await Team::genLastScore(SessionUtils::sessionTeam());
-            // Invalidate score cache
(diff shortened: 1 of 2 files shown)


justinwray:
This PR has been updated with the discussed changes and the race condition fix from #352.  This specific PR (#340) is dependent on PR #340.  You can test a full merge of all of the Memcached related PRs via [WraySec/fbctf/tree/full_merge](/WraySec/fbctf/tree/full_merge).

PR #341, opened 2016-11-05
Added Memcached Implementation to Progressive

justinwray:
* Updated Progressive::genProgressiveGameboard() to utilize Memcached.  Previously the method was calling 1(n) query per team.  Now the queries have been consolidated to a single query, and the data is stored in Memcached.  The queries will now only run after invalidation due to a change to the progressive log.

* Updated Progressive::genCount() to store the results in Memcached.  The cache is updated when the iteration changes, based on an update to the progressive log.

* The restructuring of this code has the benefit of fixing the graphic score display on the Scoreboard.

### DIFF
--- a/src/models/Progressive.php
+++ b/src/models/Progressive.php
@@ -6,8 +6,8 @@ class Progressive extends Model {
 
   protected static Map<string, string>
     $MC_KEYS = Map {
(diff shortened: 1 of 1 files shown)


justinwray:

PR #342, opened 2016-11-05
Fixed Gameboard Breakage on Team Delete

justinwray:
* Updated MultiTeam::genFirstCapture() to ensure the Gameboard doesn't break if the first team to capture a level is deleted.

### DIFF
--- a/src/models/MultiTeam.php
+++ b/src/models/MultiTeam.php
@@ -47,6 +47,10 @@ class MultiTeam extends Team {
       );
       return $mc_result;
     }
+    $all_teams = self::getMCRecords('ALL_TEAMS');
+    invariant($all_teams !== null, 'all_teams should not be null');
+    invariant($all_teams instanceof Map, 'all_teams should be of type Map');
+    return $all_teams;
   }
 
   public static async function genTeam(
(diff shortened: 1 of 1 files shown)


gsingh93:
Looks good, but what do you thinks about creating a `getMCRecords` method that doesn't return a nullable type, and instead throws an exception (you can just use `invariant`) if the key isn't in memcache? Usually the naming convention is something like `getMCRecordsX`. Then we can use that in the cases where we know we should have the key in memcache, and avoid all the duplicate `invariant` checks where we assert things aren't null.

Also, my comment about `HH_IGNORE_ERROR[4110]` applies to this PR and maybe some of the other PRs as well.

justinwray:
@gsingh93 Originally the new `getMCRecordsX` method would have made since. Now, given the new fix  within #352 which will need to be replicated to this PR (and the others) we can't definitively assert that the MC record is not null on retrieval.

However, instead the invariant calls can be condensed much like in #352 and that should clean up the duplicates.

The remaining `HH_IGNORE_ERROR[4110]` in both this PR and #339 should be solvable with a final invariant.

justinwray:

PR #343, opened 2016-11-05
Added Memcached Implementation to Link

justinwray:
* Updated Link::genAllLinks() to utilize Memcached.  Previously the method was calling 1(n) query per level.  Now the queries have been consolidated to a single query, and the data is stored in Memcached.  The queries will now only run after invalidation due to a link being altered.

### DIFF
--- a/src/models/Link.php
+++ b/src/models/Link.php
@@ -1,6 +1,12 @@
 <?hh // strict
 
 class Link extends Model {
+
+  protected static string $MC_KEY = 'links:';
+
+  protected static Map<string, string>
+    $MC_KEYS = Map {"LEVELS_COUNT" => "link_levels_count"};
+
   private function __construct(
     private int $id,
     private int $levelId,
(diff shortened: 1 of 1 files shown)


justinwray:

src/models/Link.php
```diff
@@ -45,53 +56,125 @@ public function getLink(): string {
...
+      invariant($links->contains($link_id) !== false, 'link not found');
+      if ($links->contains($link_id)) {
+        $link = $links->get($link_id);
+        invariant($link instanceof Link, 'link should be of type Link');
+        return $link;
+      }
+    } else {
+      invariant($mc_result instanceof Map, 'links should be of type Map');
+      invariant($mc_result->contains($link_id) !== false, 'link not found');
+      if ($mc_result->contains($link_id)) {
```
gsingh93: Since we have the assertion above, we don't need the if statement.

justinwray: No problem!  Will fix this now.

justinwray:

PR #344, opened 2016-11-05
Refactored Announcement Memcached Implementation

justinwray:
* Updated the Memcached implementation within the Announcement class to bring it inline with the other Memcached code.


justinwray:

PR #345, opened 2016-11-05
Added Memcached Implementation to Attachment

justinwray:
* Updated Attachment::genAllAttachments() and Attachment::genHasAttachment() to utilize Memcached.  Previously the methods were calling 1(n) queries per level.  Now the queries have been consolidated to a single query each, and the data is stored in Memcached.  The queries will now only run after invalidation due to an attachment being altered.


justinwray:

PR #346, opened 2016-11-05
Added Memcached Implementation to Category

justinwray:
* Updated the Category::genAllCategories() to utilize Memcached.  Previously this method only ran a single query but it happened frequently.  The query results are now cached and invalidated when Categories are altered.


justinwray:

justinwray:
Reopening PR which was closed by mistake

justinwray:

PR #347, opened 2016-11-05
Added Memcached Implementation to Sessions

justinwray:
* Sessions are still stored within the database; however, sessions are now replicated into Memcached.  By adding a Memcached layer on-top of Sessions, most session-related functions no longer require a query against the database, dramatically reducing the queries and improving performance.

* All session related functions are redundant; if the data is not found within the cache, but does exist in the database, it will be retrieved from the database and stored in the cache.  The redundancy ensures that failures or thrashing of Memcached do not invalidate valid sessions.

* Sessions are stored back to the database at the same frequently as they were prior to this update.  Specifically, the session is written to the database on each full-page load.

* The session cache is updated on every request, regardless of the type of request.  The continual update of the cached version of the session allows more granular updates of both 'last_access_ts' and 'last_access_page'.   The admin will always have the most up to date information available for display within the administrative Sessions page.

* As the session cache is more up-to-date, this can be used as a verification of expiration.  The old method allowed for sessions to timeout if a full page navigation did not take place within the allotted session lifetime, this is no longer the case.  If a sessions is marked as expired based on the database results, yet the cache version shows activity within the session lifetime span, the session is updated within the database and remains valid.

* Modals are specifically enforcing the various sessions updates.  The inclusion of modals in the session tracking allows for more specific 'last_access_page' values.

### DIFF
--- a/src/Router.php
+++ b/src/Router.php
@@ -13,7 +13,7 @@ class Router {
     if ($ajax) {
       return await self::genRouteAjax($page);
     } else if ($modal !== null) {
-      $xhp = await self::genRouteModel($page, strval($modal));
+      $xhp = await self::genRouteModal($page, strval($modal));
       return strval($xhp);
     } else {
       $response = await self::genRouteNormal($page);
@@ -21,7 +21,7 @@ class Router {
     }
   }
 
-  private static async function genRouteModel(
+  private static async function genRouteModal(
     string $page,
     string $modal,
   ): Awaitable<:xhp> {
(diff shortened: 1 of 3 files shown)


justinwray:

src/Router.php
```diff
@@ -25,6 +25,7 @@ class Router {
     string $page,
```
gsingh93: I noticed that the title of this function is mispelled. Can you change it to `genRouteModal`?

justinwray: Not a problem!

justinwray:
@gsingh93 The requested method name has been renamed.

Issue #351, opened 2016-11-10
Can't load Leaderboard


freeqaz:
Howdy y'all,

Last night, I finishing up some testing for a CTF and ran into some issues getting the Leaderboard working. It looks like there are some issues. Here's a screenshot.

![image](https://i.imgur.com/ucaAiYo.png)

A few of the stack traces (there were many):

```
[Thu Nov 10 12:05:12 2016] [hphp] [19:7eff6afff700:1:000001] [] \nCatchable fatal error: Value returned from async method MultiTeam::genAllTeamsCache() must be of type HH\\Map, bool given in /var/www/fbctf/src/models/MultiTeam.php on line 44
[Thu Nov 10 12:05:12 2016] [hphp] [19:7eff707ff700:7:000001] [] \nCatchable fatal error: Value returned from async method MultiTeam::genLeaderboard() must be of type array, bool given in /var/www/fbctf/src/models/MultiTeam.php on line 73
```

I'm not sure what the issue is because I don't write PHP very much. Hopefully this is an easy one to resolve.

Thanks!

-Free

gsingh93:
@justinwray This seems to be related to the MultiTeam class you added. Can you take a look?

justinwray:
@gsingh93 Roger that.  Will take a look now!

justinwray:
@freeqaz Did you happen to delete all of the teams before this issue arose?  Or did you make any changes to the teams directly in the database?

justinwray:
@gsingh93 This turned out to be a race condition, and is resolved in PR #352.

@freeqaz Thank you for this report!  This should be resolved in the pending PR referenced above.

freeqaz:
Sweet, thanks @justinwray!

PR #352, opened 2016-11-11
Fixed Memcached Race Condition

justinwray:
* Updated the usage of Memcached throughout the MultiTeam and ScoreLog classes (the only two classes utilizing the new Memcached implementation at the time) to resolve a race condition. It was possible that the cache existence could be verified and then the cache is invalidated before the cache is retrieved and utilized. Now the cache is retrieved and utilized without a second request. This resolves the race condition.

* Updated Hack/HHVM Type Checking. Removed all instance of HH_IGNORE_ERROR and updated to use invariant to confirm the type and nullability. Combined both the type and null checks into a single invariant to cleanup the code.

* Updated some code styling based on discussions on the pending PRs (relevant to these classes, and the Memcached implementation).

* This race condition fix will need to be applied to the following PRs before merge:  facebook/fbctf#339, facebook/fbctf#340, facebook/fbctf#341, facebook/fbctf#342, facebook/fbctf#343, facebook/fbctf#344, facebook/fbctf#345, facebook/fbctf#346, and facebook/fbctf#347.

* This PR is a prerequisite before the following PRs can be merged:  facebook/fbctf#339, facebook/fbctf#340, facebook/fbctf#341, facebook/fbctf#342, facebook/fbctf#343, facebook/fbctf#344, facebook/fbctf#345, facebook/fbctf#346, and facebook/fbctf#347.

* This PR resolved the following issue: facebook/fbctf#351

### DIFF
--- a/src/models/MultiTeam.php
+++ b/src/models/MultiTeam.php
@@ -6,14 +6,14 @@ class MultiTeam extends Team {
 
   protected static Map<string, string>
     $MC_KEYS = Map {
-      "ALL_TEAMS" => "all_teams",
-      "LEADERBOARD" => "leaderboard_teams",
-      "POINTS_BY_TYPE" => "points_by_type",
-      "ALL_ACTIVE_TEAMS" => "active_teams",
-      "ALL_VISIBLE_TEAMS" => "visible_teams",
-      "TEAMS_BY_LOGO" => "logo_teams",
-      "TEAMS_BY_LEVEL" => "level_teams",
-      "TEAMS_FIRST_CAP" => "capture_teams",
+      'ALL_TEAMS' => 'all_teams',
+      'LEADERBOARD' => 'leaderboard_teams',
+      'POINTS_BY_TYPE' => 'points_by_type',
+      'ALL_ACTIVE_TEAMS' => 'active_teams',
+      'ALL_VISIBLE_TEAMS' => 'visible_teams',
+      'TEAMS_BY_LOGO' => 'logo_teams',
+      'TEAMS_BY_LEVEL' => 'level_teams',
+      'TEAMS_FIRST_CAP' => 'capture_teams',
     };
 
   private static async function genTeamArrayFromDB(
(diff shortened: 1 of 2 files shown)


src/models/MultiTeam.php
```diff
@@ -35,22 +35,31 @@ class MultiTeam extends Team {
...
         $all_teams->add(
-          Pair {intval($team->get("id")), Team::teamFromRow($team)},
+          Pair {intval($team->get('id')), Team::teamFromRow($team)},
         );
       }
       self::setMCRecords('ALL_TEAMS', $all_teams);
+      return $all_teams;
+    } else {
+      invariant(
+        ($mc_result !== null && $mc_result instanceof Map),
```
gsingh93: I believe `$mc_result instanceof Map` should take care of the `$mc_result !== null` case, but you can double check that. Same for any other place where you do this.

src/models/ScoreLog.php
```diff
@@ -77,46 +83,87 @@ private static function scorelogFromRow(Map<string, string> $row): ScoreLog {
...
-        /* HH_IGNORE_ERROR[4062] */
-        return $level_captures->get($level_id)->linearSearch($team_id) != -1;
+        $level_capture_teams = $mc_result->get($level_id);
+        invariant(
+          ($level_capture_teams !== null &&
+           $level_capture_teams instanceof Vector),
+          'level_capture_teams should of type Vector and not null',
+        );
+        $team_id_key = $level_capture_teams->linearSearch($team_id);
+        return $team_id_key != -1;
```
gsingh93: Always use `!==`

src/models/ScoreLog.php
```diff
@@ -77,46 +83,87 @@ private static function scorelogFromRow(Map<string, string> $row): ScoreLog {
...
-        $team_id_key = /* HH_IGNORE_ERROR[4062]: getMCRecords returns a 'mixed' type, HHVM is unsure of the type at this point */
-          $level_captures->get($level_id)->linearSearch($team_id);
+        $level_capture_teams = $mc_result->get($level_id);
+        invariant(
+          ($level_capture_teams !== null &&
+           $level_capture_teams instanceof Vector),
+          'level_capture_teams should of type Vector and not null',
+        );
+        $team_id_key = $level_capture_teams->linearSearch($team_id);
         if ($team_id_key != -1) {
```
gsingh93: `!==`

justinwray:
Updated based on feedback!

gsingh93:
Thanks for the fix!

Issue #358, opened 2016-11-14
Fatal error: Uncaught exception (country.php)


jeszicawii:
Hi all,

Have had this issue a number of times, I have to re-provision just about every time, any guidance on this would be great.

When loading the map it sticks at 100% loading and does nothing.

[Mon Nov 14 12:56:31 2016] [hphp] [948:7f61887ff700:1685:000001] [] \nFatal error: Uncaught exception 'HH\\InvariantException' with message 'Expected exactly one result' in /var/www/fbctf/src/models/Country.php:207\nStack trace:\n#0 /var/www/fbctf/src/models/Country.php(207): HH\\invariant_violation()\n#1 /var/www/fbctf/src/inc/gameboard/listview.php(17): Country::gen()\n#2 (): ListviewController->genRender()\n#3 (): HH\\WaitHandle->join()\n#4 /var/www/fbctf/src/inc/gameboard/listview.php(49): HH\\Asio\\join()\n#5 {main}

iangcarroll:
How many countries are in your `countries` table (`select count(*) from countries`)? Not sure why this wouldn't be populated.

gsingh93:
@justinwray do you know if this is related to any of the recent memcache changes?

justinwray:
@gsingh93 It doesn't appear so.  The updates to the `Country` class Memcached implementation are in PR #339 (not yet merged).  In this case it's a straight MySQL Query:

```
  // Get a country by id.
  public static async function gen(int $country_id): Awaitable<Country> {
    $db = await self::genDb();

    $result = await $db->queryf(
      'SELECT * FROM countries WHERE id = %d LIMIT 1',
      $country_id,
    );

    invariant($result->numRows() === 1, 'Expected exactly one result');
    $new_country =
      await self::countryFromRow(firstx($result->mapRows())->toArray());
    return $new_country;
  }
```

So other than the `countries` table being empty, the only other logical cause of the violation is an invalid `$country_id` being passed.  In fact, this actually seems to be the issue, within `ListviewController` class we find this:

`$country = await Country::gen(intval($level->getId()));`

Which it seems should be:

`$country = await Country::gen(intval($level->getEntityId()));`

justinwray:
@gsingh93 I just created and tested PR #359 which fixes this issue.  I updated ListviewController::genRender() to pass the correct ID.

gsingh93:
Good catch! I merged the PR, so this should be fixed.

PR #359, opened 2016-11-15
Fixed Issue #358

justinwray:
* Updated ListviewController::genRender() to pass the Entity (Country) ID instead of the Level ID to Country::gen().


gsingh93:
Thanks!

javuto:
Awesome, thank you @justinwray

Issue #234, opened 2016-09-26
"502 Bad Gateway" NGINX error when HHVM does not start on virtual machine restart


eeeehaw:
I thought I had licked this 502 Bad Gateway error when running FBCTF on Azure, after changing the private internal subnet address space to 10.10.10.0/24 and setting the virtual machine (VM) eth0 IPaddr to 10.10.10.5 (see Closed Issue https://github.com/facebook/fbctf/issues/219 ).  But then I discovered that after restarting the VM, the problem came back.  I then tried Google Cloud and saw the exact same problem occur on that VM, which uses a different hyper-visor.  Of course, both VMs are running Ubuntu 14.04LTS.  'sudo service hhvm status' shows hhvm running after initial provisioning, but not running after the VM is restarted.

I am able to correct the problem temporarily by re-provisioning FBCTF again ('./extra/provision.sh -m prod -s $PWD'); then everything is back to normal...until the VM is restarted again.  Meanwhile, mysql & nginx services restart properly and run even after hhvm fails to start.  As a volunteer for a high school  cyber security club (= no money aside of volunteer out-of-pocket), I don't want a Azure or Google Cloud VM running when not needed so as to keep my out-of-pocket costs down, thus the need to have those environments stopped when not needed and restarted when needed by admin or students.

I have posted this problem in the HHVM project Issues page also.  A search on NGINX did not reveal any relevant causes as far as I could determine (altho they emphasized file permissions as a frequent cause).  Could there be a permissions change that occurs during a restart that may be affecting sockets or something else, which I presume would fall into the FBCTF area scripts?

iangcarroll:
I cannot say why setting that IP resolved your problem, because the source code does not use that IP, Vagrant does, but either way:

Try running the following after rebooting:
```
sudo rm /var/run/hhvm/hhvm.hhbc
sudo hhvm-repo-mode enable "/var/www/fbctf"
sudo chown www-data:www-data /var/run/hhvm/hhvm.hhbc
sudo service hhvm start
```

eeeehaw:
Ian, that does work to recover HHVM to running service, and resolve the 
nginx 502 Bad Gateway error.  Helpful as a clue toward finding root 
cause of the problem?

Here's my console output for that:
----------------
eeeehaw@cyberwars-vm1:~$ sudo rm /var/run/hhvm/hhvm.hhbc
eeeehaw@cyberwars-vm1:~$ sudo hhvm-repo-mode enable "/var/www/fbctf"
[...]

PR #367, opened 2016-11-21
HHVM Repo Authoritative mode optional and change provision arguments

javuto:
Mitigation for the ```502 Bad Gateway``` problem (https://github.com/facebook/fbctf/issues/234) when restarting the FBCTF machine, making HHVM Repo Authoritative mode optional.
Also, providing the environment variable ```FBCTF_PROVISION_ARGS```, the arguments to provision can be modified for the vagrant machine.


javuto:
For example, if you want to deploy the platform in production mode, you can use the command:
```FBCTF_PROVISION_ARGS='-m prod -s /vagrant -d /var/www/fbctf' vagrant up```
And there is no need to modify the ```Vagrantfile```

gsingh93:
This would mean we slow down production deployments significantly. I think we should figure out what the actual problem is instead of just hiding it.

PR #371, opened 2016-11-28
Updated Session Invalidation

justinwray:
* The new Session::getUnprotectedSessions() method returns all non-admin sessions and is now used in conjunction with the new Session::genDeleteAllUnprotected() method.  When an admin disables all teams, all non-administrative sessions are deleted and invalidated.

* Session::genDeleteByTeam is called to delete and invalidate team sessions on password update, team status toggle, and team deletion.

* Added condition to Configuration::genUpdate() to delete and invalidate all user sessions when an administrator disables login via the ‘login off’ admin toggle.

* The SQL limit was removed from Session::genDeleteByTeam() delete query. This resolves the issue where only one team session is being deleted after a team deletion.

* Updated the Team class to add session invalidation when team settings are changed or updated.  This includes invalidation of all related sessions when: the admin status is changed for a team, an individual team is disabled, all teams are disabled, a team’s password is updated, or a  team is deleted.

### DIFF
--- a/src/models/Configuration.php
+++ b/src/models/Configuration.php
@@ -58,6 +58,9 @@ class Configuration extends Model {
       $value,
       $field,
     );
+    if ($field === 'login' && intval($value) === 0) {
+      await Session::genDeleteAllUnprotected();
+    }
 
     self::getMc()->delete(self::MC_KEY.$field);
   }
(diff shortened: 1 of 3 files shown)


Mario1120:
No one pay attention to this one... why?

gsingh93:
Merged, thanks.

PR #376, opened 2016-12-02
Fix for HHVM not started on reboot

javuto:
Fix for https://github.com/facebook/fbctf/issues/234 that was fixed by https://github.com/facebook/hhvm/issues/7504

Now when provisioning in production mode, and therefore, with HHVM Repo Authoritative mode enabled, HHVM will start after a reboot.


gsingh93:
What are the permissions on `/var/cache/hhvm`? Are they the same as `/var/run/hhvm`? We want to make sure an unprivileged user can't put their own bytecode there.

javuto:
I think it should be fine:
```
-rw-r--r--  1 www-data www-data  19M Dec  2 00:46 hhvm.hhbc
```

gsingh93:
Works for me.

PR #379, opened 2016-12-03
Updated PR #259 - Fix progressive scoreboard and bases

justinwray:
* Merged PR #259 and PR #264

* Updated the included classes within bases.php, adding the Link, Team, ScoreLog, Control, and MultiTeam classes.

* Added scoring invalidation on Hint usage, which will ensure the scoreboard is updated on any event that alters team scores.

* Updated the formatting of the new code.

### DIFF
--- a/src/scripts/bases.php
+++ b/src/scripts/bases.php
@@ -1,11 +1,12 @@
 <?hh
 
-if (php_sapi_name() !== 'cli') {
-  http_response_code(405); // method not allowed
-  exit(0);
-}
-
-require_once ('/var/www/fbctf/vendor/autoload.php');
+require_once('/var/www/fbctf/src/Db.php');
+require_once('/var/www/fbctf/src/Utils.php');
+require_once('/var/www/fbctf/src/models/Model.php');
+require_once('/var/www/fbctf/src/models/Importable.php');
+require_once('/var/www/fbctf/src/models/Exportable.php');
+require_once('/var/www/fbctf/src/models/Level.php');
+require_once('/var/www/fbctf/src/models/Configuration.php');
 
 $conf_game = \HH\Asio\join(Configuration::gen('game'));
 while ($conf_game->getValue() === '1') {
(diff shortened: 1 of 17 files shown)


src/scripts/bases.php
```diff
@@ -5,7 +5,18 @@
   exit(0);
 }
 
-require_once ('/var/www/fbctf/vendor/autoload.php');
+require_once ('/var/www/fbctf/src/Db.php');
```
javuto: I think we better use:
```
require_once (__DIR__.'/../Db.php');``` 
require_once (__DIR__.'/../Utils.php');
require_once (__DIR__.'/../Model.php');
require_once (__DIR__.'/../Importable.php');
require_once (__DIR__.'/../Exportable.php');
require_once (__DIR__.'/../Level.php');
require_once (__DIR__.'/../Link.php');
require_once (__DIR__.'/../Team.php');
require_once (__DIR__.'/../Configuration.php');
require_once (__DIR__.'/../ScoreLog.php');
require_once (__DIR__.'/../Control.php');
require_once (__DIR__.'/../MultiTeam.php');
```
to avoid the hardcoded path, and I think that is the main problem now.

javuto:
We may want to find a better way to do this: https://github.com/facebook/fbctf/blob/master/src/models/Progressive.php#L150 since that location has changed for the main ```.hhbc``` file (https://github.com/facebook/fbctf/blob/master/extra/hhvm.conf#L22). In any case, it may fail if repo mode is not enabled.
And the same with bases: https://github.com/facebook/fbctf/blob/master/src/models/Level.php#L1127

src/scripts/bases.php
justinwray: Good catch!  Committing the change now.

src/scripts/progressive.php
```diff
@@ -4,8 +4,11 @@
   http_response_code(405); // method not allowed
   exit(0);
 }
-
-require_once ('/var/www/fbctf/vendor/autoload.php');
```
justinwray: No problem!

justinwray:
Updated the paths based on the feedback from @javuto.

javuto:
Looks good to me! @gsingh93?

gsingh93:
Yup, merged.

PR #385, opened 2016-12-05
Small changes to provisioning

javuto:


extra/hhvm.conf
```diff
@@ -19,7 +19,7 @@ hhvm.log.header = true
 hhvm.log.access[default][file] = /var/log/hhvm/access.log
 hhvm.log.access[default][format] = "%h %l %u %t \"%r\" %>s %b"
 hhvm.server.source_root = CTFPATH
-hhvm.repo.central.path = /var/cache/hhvm/hhvm.hhbc
+hhvm.repo.central.path = /var/run/hhvm/hhvm.hhbc
```
javuto: `/var/run/hhvm/hhvm.hhbc` is created when the service starts and can be accessed by www-data without having to touch ownership or permissions. But `/var/cache/hhvm/hhvm.hhbc` will require to adjust permissions and the point is to persist after reboots, which the files under `/var/run` do not.

gsingh93: But you call `sudo chown www-data:www-data "$__repofile"` regardless of which one we're using. It seems like more trouble to allow for both with no benefit.

PR #390, opened 2016-12-07
Set default values for columns in `teams` table

JasonTarka:
SQL Strict Mode requires a value to be set for `NOT NULL` columns, and throws an
error if a NULL value is attempted to be inserted, or the value is not specified
in an `INSERT` statement and no default is set.

The creation of the admin user during the provision script does not specify the
`points` column. The creation of teams through the UI additionally specifies
neither the `admin` nor `protected` columns. Since these columns are all
`NOT NULL` MySQL throws an error when running the `INSERT`.

Adding default values to the columns fixes this issue, and adds extra safety by
ensuring a team is not administrators or protected if the columns are omitted.


gsingh93:
Thanks!

PR #392, opened 2016-12-09
Added Game Schedule Functionality

justinwray:
fbctf_game_schedule

* Allow an administrator to set the start and end date/time for the game.

* If the game has not started and is set to start in the future the "countdown" page will display the date of the upcoming game and the time remaining until the game begins.

* If the game is currently running the "countdown" page will display "In Progress."

* If no game is schedule the "countdown" page will display the default of "Soon."

* The administrator can change the end date/time of the game at any point and the various timers/displays will update appropriately.

* Configuration table within the database has been updated to define the data type found in the "next_game" field.

* JavaScript have been updated to handle the submission and processing of the new schedule fields on the admin Configuration page.

* Note: This PR does NOT perform any automation of the game starting or stopping, instead simply providing scheduling capability which will impact the various timers.

### DIFF
--- a/src/controllers/AdminController.php
+++ b/src/controllers/AdminController.php
@@ -272,6 +272,7 @@ class AdminController extends Controller {
 
   public async function genRenderConfigurationContent(): Awaitable<:xhp> {
     $awaitables = Map {
+      'game' => Configuration::gen('game'),
       'registration' => Configuration::gen('registration'),
       'registration_players' => Configuration::gen('registration_players'),
       'login' => Configuration::gen('login'),
@@ -295,6 +296,7 @@ class AdminController extends Controller {
 
     $results = await \HH\Asio\m($awaitables);
 
+    $game = $results['game'];
     $registration = $results['registration'];
     $registration_players = $results['registration_players'];
     $login = $results['login'];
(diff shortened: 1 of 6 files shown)


CJHackerz:
How to use this?

justinwray:
Within the Administration page on the Configuration tab you will now find the game start and end date/time (as shown in the attached GIF above). You just need to set those values for the various timers and displays to update.

javuto:
Looks good to me!

wisco24:
Yes Please!  Thanks!   When will this get merged?

src/controllers/AdminController.php
```diff
@@ -334,12 +336,45 @@ class="fb-cta cta--yellow"
     $timer_on = $timer->getValue() === '1';
     $timer_off = $timer->getValue() === '0';
 
-    if ($start_ts->getValue() === '0') {
-      $start_ts = tr('Not started yet');
-      $end_ts = tr('Not started yet');
+    if ($start_ts->getValue() !== '0' && $start_ts->getValue() !== 'NaN') {
```
iangcarroll: Can something actually get returned as the string literal `NaN`?

justinwray: Yes, "NaN" (not a number) can be returned via the client (JavaScript). However, this should not happen with the current client-side code. It was included as a fail-safe.

src/controllers/AdminController.php
```diff
@@ -334,12 +336,45 @@ class="fb-cta cta--yellow"
...
+      $game_start_ts = $start_ts->getValue();
```
justinwray: I'm not sure which section of code you thought would be better as an array.  I am likely just missing it.  Do you mind clarifying?

iangcarroll: If XHP supports it, then the game_start variables; something like:
```php
$game_start = [
     'year' => '0',
     'month' => '0',
     'day' => '0',
     // etc
];
```

There is also `getdate()`, which should return all the data you want as an array.

justinwray: Updated to use getdate().  Thanks @iangcarroll!

justinwray:
This PR has been updated based on the feedback.

PR #394, opened 2016-12-09
Force Attachment Permissions

justinwray:
* Set uploaded attachment permissions to 0600.

### DIFF
--- a/src/models/Attachment.php
+++ b/src/models/Attachment.php
@@ -68,17 +68,12 @@ class Attachment extends Model {
       );
 
       // Force 0600 Permissions
-      chmod(must_have_string($server, 'DOCUMENT_ROOT').$local_filename, 0600);
+      $chmod = chmod(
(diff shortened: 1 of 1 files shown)


gsingh93:
Just to be super safe, can you assert that the result of the chmod is `true`?

wrayjustin:
Updated with validation of the permission change.

justinwray:
Updated with validation of the permissions.

gsingh93:
`chmod` returns true or false depending on whether it succeeded or failed in changing the permissions. Is there any benefit to the assertion you added as opposed to just asserting the return value is `true`?

justinwray:
The implicit permission check was added for extra insurance of the file permissions.  However, there doesn't seem to be much benefit as the chmod() function should always provide a factual return (that is, there are not any documented cases where the chmod return is inaccurate).

I've updated the code to verify the return.

PR #419, opened 2017-01-19
Database replication

justinwray:
Database Replication/Load Balancing

- Initial database replication pull request for discussion and review.  A few technical items to consider:

- The implementation uses a master/master configuration versus a master/slave configuration due to incompatibilities with HHVM and mysqlnd.

- The implementation supports Vagrant and direct provisioning only at this time.  Vagrant users should utilize the included install_config.bat script to rebuild their vagrant configuration.  Those provisioning directly can execute the provision script on the initial host and the replication script on the subsequent database servers.

- When directly provisioning the user needs to supply the replication options and the server number for which they are deploying (starting from one, onwards).  Vagrant will automatically handle this process based on the options provided to the install_config.bat script.

- The install_config.bat file can be extended to include further configuration of the Vagrant options.  Such options might include a different network subnet, alternative usernames, memory allocation, etc.  Furthermore, this script can be duplicated as a bash file for Linux-based host systems.

- The current implementation of Memcached provides significant optimization to the database queries and subsequent caching.  Database replication is unlikely to provide benefit to typical users and should only be used for complex environments.

- The settings.ini file will need to be updated to include the DB_MASTER and DB_SLAVE fields.  DB_MASTER should be the primary server (web interface host), and DB_SLAVE should be your secondary MySQL server.  Additionally the database username and password within settings.ini needs to be updated to the replicator account.


juliannagler:
More context on alternative approach for discussion, from @justinwray 

"Given the complications and inefficiency of this process, we believe it would be best to support replication within the Hack code base and defer to the end-user to configure the subsequent database servers and the replication. Documentation can be provided for the setup process, which is fairly straightforward. Once the secondary database systems are configured, the end-user could update the relevant settings in the settings.ini file or we could develop an interface within the project that allows them to define the replication servers. "

PR #420, opened 2017-01-21
Disable Country Input When Scoring Disabled

justinwray:
* When scoring is disabled a message will now be displayed to the user when selecting a country.

* The input box for the country is disabled.

* The hint button is removed.

* The text of the submit button is changed to indicate that scoring is off and the button does not perform any action.


javuto:
Great job @justinwray! This is something that was needed for a while.

juliannagler:
Wahoo! This is one of the most common feedback items we get.

Issue #319, opened 2016-11-01
Allow the admins to log in if login is disabled


gsingh93:
If login is disabled, the admin account is not allowed to log in if they're logged out. We should always allow the admins to log in.

PR #423, opened 2017-01-22
Allow Admin Login When Team Login Disabled

justinwray:
* Added an "Admin Login" button to the disabled login page.

login_disabled

* Only valid admin accounts can login when team login is disabled.

* Provide notice to users that only an admin can login at the time (when team login is disabled).

admin_login

### DIFF
--- a/src/controllers/ajax/IndexAjaxController.php
+++ b/src/controllers/ajax/IndexAjaxController.php
@@ -226,7 +226,7 @@ class IndexAjaxController extends AjaxController {
     // Check if login is disabled and this isn't an admin
     $login = await Configuration::gen('login');
     if (($login->getValue() === '0') &&
-        (!$team || (($team) && ($team->getAdmin() === false)))) {
+        ($team === null || $team->getAdmin() === false)) {
       return Utils::error_response('Login failed', 'login');
     }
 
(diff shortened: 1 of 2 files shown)


javuto:
This PR implements https://github.com/facebook/fbctf/issues/319

src/controllers/ajax/IndexAjaxController.php
```diff
@@ -220,15 +220,17 @@ protected function getActions(): array<string> {
...
   ): Awaitable<string> {
-    // Check if login is enabled
+    // Verify credentials first so we can allow admins to login regardless of the login setting
+    $team = await Team::genVerifyCredentials($team_id, $password);
+
+    // Check if login is disabled and this isn't an admin
     $login = await Configuration::gen('login');
-    if ($login->getValue() === '0') {
+    if (($login->getValue() === '0') &&
+        (!$team || (($team) && ($team->getAdmin() === false)))) {
```
gsingh93: You can get rid of the redundant `($team)` I think. You should just be able to do: `$team === null || $team->getAdmin() === false`. I also changed `!$team` to an explicit null check, because I don't like mixing bools with nullable values.

justinwray:
@gsingh93 Updated the condition block based on your feedback.

PR #424, opened 2017-01-22
Fix fail error

shargon:
Throw this error:

> [Sun Jan 22 20:18:19 2017] [hphp] [13528:7f3ebf7ff700:4:000001] [] \nNotice: Undefined variable: genCheckStatus in /var/www/fbctf/src/controllers/AdminController.php on line 2702
> 
> [Sun Jan 22 20:18:19 2017] [hphp] [13528:7f3ebf7ff700:4:000002] [] \nFatal error: Function name must be a string in /var/www/fbctf/src/controllers/AdminController.php on line 2702


javuto:
Good catch! Gracias 👍

shargon:
Thanks you for your project!

PR #427, opened 2017-01-27
Flush Memcached and Reset Database Admin Utilities

justinwray:
* Allow administrators to flush Memcached directly from the interface.  Flushing Memcached is required anytime data is manually manipulated within the database, additionally, if there is any caching issue, flushing Memcached is recommended.  The cache will rebuild after a flush without any additional action required by the user.

* Allow administrators to reset the database.  Resetting the database is useful when wanting to wipe all data from the application. Additionally, the process can be used to upgrade to the latest database schema.  Resetting the database will remove all current data (except admin users) and flush the data from the cache.  Administrative users will remain after a database reset.

* Added new "Utilities" line to the admin Control page.  Both the Flush Memcached and Reset Database buttons have been added to this section.

screen shot 2017-01-27 at 12 27 56 pm

* The database reset button will prompt the user before resetting the database to ensure they understand the action will erase all existing data.

screen shot 2017-01-27 at 12 28 06 pm

### DIFF
--- a/src/controllers/AdminController.php
+++ b/src/controllers/AdminController.php
@@ -913,6 +913,33 @@ class AdminController extends Controller {
           </section>
           <section class="admin-box">
             <header class="admin-box-header">
+              <h3>{tr('Utilities')}</h3>
+            </header>
+            <div class="fb-column-container">
+              <div class="col col-pad col-1-4">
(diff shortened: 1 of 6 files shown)


javuto:
Looks good, and flushing the keys was a much needed feature!

PR #433, opened 2017-01-31
Fix previous fail (autofocus dosen't work)

shargon:


javuto:
This does not fix xhp for autofocus, the solution requires `autofocus="on"`

PR #436, opened 2017-02-01
Fix autofocus error

shargon:
Sorry for that


javuto:
Already fixed in https://github.com/facebook/fbctf/commit/3891d005072b65d925137c556fac4b0687132223

PR #439, opened 2017-02-03
Registration enforcing strong passwords

javuto:
Strongs passwords now can be enforced, with a configuration radio button and a dropdown select to pick how strong the password will be:

image

The registration window will look like this when the chosen password is not strong enough:

image


javuto:
Closing this PR to start developing in the `dev` branch instead.

PR #442, opened 2017-02-03
Registration enforcing strong passwords

javuto:
Strongs passwords now can be enforced, with a configuration radio button and a dropdown select to pick how strong the password will be:

image

The registration window will look like this when the chosen password is not strong enough:

image

### DIFF
--- a/src/controllers/IndexController.php
+++ b/src/controllers/IndexController.php
@@ -470,7 +470,7 @@ class IndexController extends Controller {
               <h6>{tr('Choose an Emblem')}</h6>
               <h6>
                 <a href="#" id="custom-emblem-link">
-                  {tr('or Upload your own')}
+                  {tr('or upload your own')}
                 </a>
               </h6>
               <div class="custom-emblem">
--- a/src/controllers/IndexController.php
+++ b/src/controllers/IndexController.php
@@ -345,6 +345,7 @@ class IndexController extends Controller {
                   name="teamname"
                   type="text"
                   maxlength={20}
+                  autofocus={true}
                 />
                 {$ldap_domain_suffix}
               </div>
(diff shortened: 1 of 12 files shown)


src/controllers/IndexController.php
```diff
@@ -451,20 +452,25 @@ class="fb-main page--registration full-height fb-scroll">
...
+                <h6 style="color:red;">{tr('Password is too simple')}</h6>
+              </div>
               {$token_field}
             </fieldset>
             <div class="fb-choose-emblem">
               <h6>{tr('Choose an Emblem')}</h6>
               <h6>
                 <a href="#" id="custom-emblem-link">
-                  {tr('or upload your own')}
+                  {tr('or Upload your own')}
```
gsingh93: What's with this capitalization?

javuto: Good catch. Maybe we can uppercase the checks?

javuto:
The schema for the testing database was outdated, now all the tests are passing:
```
vagrant@facebookctf-dev:/var/www/fbctf$ ./extra/run_tests.sh /var/www/fbctf/
[+] Starting tests setup in /var/www/fbctf/
[+] DB Connection file
[+] Starting tests
PHPUnit 5.3.2 by Sebastian Bergmann and contributors.

................................................................. 65 / 66 ( 98%)
.                                                                 66 / 66 (100%)

Time: 7.4 seconds, Memory: 29.60Mb

OK (66 tests, 362 assertions)
WARNING: Logging before InitGoogleLogging() is written to STDERR
E0206 08:08:26.384719 20753 Singleton-inl.h:148] Singleton of type N8facebook6common12mysql_client16AsyncMysqlClientE has a living reference at destroyInstances time; beware! Raw pointer is 0x7fd39218ec00. It is very likely that some other singleton is holding a shared_ptr to it. Make sure dependencies between these singletons are properly defined.
E0206 08:08:26.403004 20753 Singleton.cpp:54] Singletons of the following types had living references after destroyInstances was finished:
	N8facebook6common12mysql_client16AsyncMysqlClientE
beware! It is very likely that those singleton instances are leaked.
[+] Deleting test database
```

PR #446, opened 2017-02-09
I think its bad

shargon:
For security reasons, i think its very bad hardcode the admin team_id by default when creating a new session row.


wisco24:
Whats your use case?   Attackers can only get this if they have access to the DB or the webapp, and if they have that, they can also just reset the admin password.

gsingh93:
This isn't a security risk, but we might as well change it.

Issue #203, opened 2016-08-31
Sponsors brands


acjcool:
We have some sponsors for our tournament. Is there no way that we can add the sponsors brands/logo's on the app? For instance, this tournament is sponsored by 'x' and then their logo's can scroll at the top of the screen or something like that? For instance, gold sponsors = xxx, silver sponsors = xxx etc. And if someone clicks on a sponsors logo then it should direct them to the sponsor page. Some sponsors have discounts to courses, products etc specifically for the tournament so it could direct them to that special discount page. Also, can we improve the leadboard screen so that it auto refreshes with the sponsors details so that we can leave it on the big screen for all to see the progress? And if a team claims first spot, then some flashing on the leaderboard would be nice to show that there was a change in the rankings.

acjcool:
Hi,

This is what I have done so far but it would be a lot better to have it in the configurations menu instead of a hack attempt in the Controller.php file:

<?hh // strict

  echo "<html>";
  echo "<body>";
  echo "<marquee> My FB CTF Competition. Gold Sponsors: Sponsor 1, Sponsor 2. Silver Sponsors: Sponsor 3, Sponsor 4. etc</marquee>";
  echo "</body>";
  echo "</html";

  abstract class Controller {
  abstract protected function getTitle(): string;
  abstract protected function getFilters(): array<string, mixed>;
  abstract protected function getPages(): array<string>;

.....

acjcool:
I've added the code in the code bracket this time as I noticed that the html tags have been removed for obvious reasons

`  echo "<html>";
  echo "<body>";
  echo "<marquee> My FB CTF Competition. Gold Sponsors: Sponsor 1, Sponsor 2. Silver Sponsors: Sponsor 3, Sponsor 4. etc</marquee>";
  echo "</body>";
  echo "</html";`

javuto:
Merged in https://github.com/facebook/fbctf/pull/448

Issue #93, opened 2016-05-18
Ability to custom brand the CTF with your name and logo.


tbowman01:
Could you provide an interface in the admin menu to brand the CTF?
ie.  upload and use a custom logo/icon and change the "Powered by Facebook" 

Temporary work around in order to customize/brand the CTF the logo and file can be found here:
- Logo Location: /fbctf/src/static/svg/icons/social-facebook.svg
- File Location:  /fbctf/src/xhp/Fbbranding.php

If you place your own logo in the same location and update the code you could start to customize the feel.  The /fbctf/src/static/css/scss/_navigation-branding.scss file has the style sheet code to customize the site further.

mehnomy:
This is a non commercial license agreement when using the Platform which has been created and released by Facebook, i don't agree on having a way to change it from being Branded with Facebook since they've have created this amazing platform.

> Facebook CTF retains a Creative Commons (CC) License, which means it can be used by any non-commercial entity for **educational purposes**. The platform was built with education and competition organizers in mind. Dozens of high school and college-level student clubs, national competition organizations, educational institutions, and industry groups have already run successful competitions using the Facebook CTF platform.

When creating an event e.g at a University for instance students would be more engaged and willing to participate seeing the Facebook logo over a unknown logo, and if you were to have a landing page for information then i think thats more then enough.

The site looks amazing as it is, if you really want to customise or if anyone else wanted to change the feel i don't really see the need of providing a admin editor to do such things, when it all can be done via the css.


Kind Regards,
Byron Fisher

tbowman01:
As the intent is for educational purposes with the education and competition organizations in mind.  I interpret this as internal education/competitions among teams.  The suggestion was geared more towards those high school or less tech savvy who do not know where to find the cascading style sheets or how to read where they are linking from.  

For the high school group or possibly even college level the "Powered by Facebook" may have some appeal.  However, in industry, as I have pitched using the platform to management for internal training.  I have recieved multiple comments about Facebook.  That initially is a challenge and may be best addressed with stating that the Facebook CTF retains a Creative Commons (CC) License, which means it can be used by any non-commercial entity for educational purposes. The platform was built with education and competition organizers in mind.

I agree the platform looks great.  I also see some room for growth and possibly forks of it as occasionally happens with open source.  If there is a way to upload multiple questions I can see a possible test bank of questions for the jeopardy style.

[...]

PR #448, opened 2017-02-11
Custom branding for icon and text

javuto:
This pull request implements the customization for brand, which has been requested in https://github.com/facebook/fbctf/issues/203 and https://github.com/facebook/fbctf/issues/93. It allows to customize the logo and the text underneath. By default it is turned off and the displayed text is "Powered By Facebook", with the small FB logo. The new settings are under the "Branding" section:

screen shot 2017-02-11 at 04 34 22

screen shot 2017-02-11 at 04 25 54

When enabling the option for a "Custom Logo", the logo can be uploaded, and it will be displayed. The text can be edited at any moment:

screen shot 2017-02-11 at 04 26 32

Anywhere in the game where the branding is displayed, will have the custom view (Gameboard, Registration, View Mode):

screen shot 2017-02-11 at 04 26 40

### DIFF
--- a/src/xhp/Custombranding.php
+++ b/src/xhp/Custombranding.php
@@ -2,17 +2,18 @@
 
 class :custombranding extends :x:element {
   category %flow;
+  attribute
+    string brandingText,
+    string brandingLogo;
 
   protected string $tagName = 'custombranding';
 
   protected function render(): XHPRoot {
-    $custom_text = \HH\Asio\join(Configuration::gen('custom_text'));
-    $custom_image = \HH\Asio\join(Configuration::gen('custom_logo_image'));
     return
       <span class="branding-el">
-        <img class="icon-badge" src={strval($custom_image->getValue())}/>
+        <img class="icon-badge" src={$this->:brandingLogo}/>
         <br/>
-        <span class="icon-text">{tr(strval($custom_text->getValue()))}</span>
+        <span class="icon-text">{$this->:brandingText}</span>
       </span>;
   }
 }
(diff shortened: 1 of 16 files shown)


stevcoll:
Testing looks good to me, nice work!

src/xhp/Custombranding.php
```diff
@@ -0,0 +1,18 @@
+<?hh // strict
+
+class :custombranding extends :x:element {
+  category %flow;
+
+  protected string $tagName = 'custombranding';
+
+  protected function render(): XHPRoot {
+    $custom_text = \HH\Asio\join(Configuration::gen('custom_text'));
```
gsingh93: I think it'd make more sense to move this out of this XHP element, and pass it in through an attribute. In general, we shouldn't be introducing more `join`s into the code, only `await`.

src/xhp/Custombranding.php
```diff
@@ -0,0 +1,18 @@
...
+    $custom_image = \HH\Asio\join(Configuration::gen('custom_logo_image'));
```
gsingh93: I think it'd make more sense to move this out of this XHP element, and pass it in through an attribute. In general, we shouldn't be introducing more `join`s into the code, only `await`.

javuto:
I replaced the `join` calls in the `xhp` branding files by attributes and also I put the `genRenderBranding` method in the main `Controller.php` file, so it is reused across admin, game and view mode controllers.

ciscomk:
@gsingh93 - is there a reason for not merging this into the master branch?

gsingh93:
We wanted to merge all new stuff to dev first, let it sit for a bit, and then merge to master. The problem now is that there are merge conflicts.

cc: @javuto

Issue #10, opened 2016-05-10
Support starting a game at a set time


gsingh93:
At the moment you have to click "Begin Game" manually to start a game. We should add a feature to set the start time and a duration or end time, and have the game automatically start at the start time.

dylanjacob:
👍

constance628:
Where to click "Begin Game" manually to start a game?
TIMER of Configuration only GAME DURATION can be set, did I miss something?

mpomarole:
@constance628 you want to be clicking on the button highlighted to being game:
<img width="1396" alt="screen shot 2016-05-14 at 19 52 30" src="https://cloud.githubusercontent.com/assets/333924/15270166/7d4fcd92-1a0d-11e6-919a-e936024e39c1.png">

gsingh93:
@javuto should we move that to "Controls" instead?

constance628:
@mpomarole Thank you, I didn't notice that there is a scroll bar beside main menu....

dylanjacob:
But what about setting a time in the future for the game to automatically start without having to have a person actually click "Begin Game"?  If I wanted to schedule for, say 9:00am tomorrow, and then  allow people to register today, how would I accomplish that?

My suggestion would be to have a radio on/off for "Start: immediately/future" and then when future is selected show a date/time picker for the future date/time.  Once that is entered and saved, the clock that appears when you click play would count down to 9:00am tomorrow and the page would refresh at 0 to allow users who have already logged in to enter the game at almost the same time.

gsingh93:
> But what about setting a time in the future for the game to automatically start without having to have a person actually click "Begin Game"?

That's what this issue is for.

colbyjoe1919:
We can't play these games people I know it's a long shift but the end
processes are going to invoke torture and collect data themselves.  We have
a dark Lord to satisfy.  Let get serious here.

juliannagler:
Relevant to #392 - automation is coming

juliannagler:
Addressed in #449

PR #449, opened 2017-02-12
Automated Game Start and Stop

justinwray:
* Games will automatically start and stop at their scheduled times.  Administrators can still manually start or stop a game regardless of the configured schedule.

* Both Control::genAutoBegin() and Control::genAutoEnd() were added to check the current time against the scheduled start or stop time and perform the relevant action (Control::genBegin or Control::getEnd).

* Control::genAutoRun() checks the current game status and determine if the game should be starting or ending, calling the appropriate function (Control::genAutoBegin or Control::getAutoEnd) and is exclusively used in the new autorun.php script.

* Control::genRunAutoRunScript() runs the new autorun.php script, ensuring the script is not already running before starting a new copy.

* The Router class was updated to include a call to Control::genRunAutoRunScript(), this ensures the script is always running.  This script status check, and execution when needed, only takes place on a full page load.

* The autorun.php script runs Control::genAutoRun() and sleeps up to 30 seconds.

* If the game is scheduled to start or stop within 30 seconds, the script will sleep for the necessary amount of time.

* Games will always start with at most a 29-second difference from the scheduled time.  This discrepancy can only take place if the schedule is changed within 30 seconds of the previously scheduled time.  Otherwise, the execution will happen at the scheduled time.

* This automation is self-contained and requires no additional dependencies or external services (like cron, etc.).

### DIFF
--- a/src/models/Control.php
+++ b/src/models/Control.php
@@ -271,11 +271,14 @@ class Control extends Model {
   public static async function genRunAutoRunScript(): Awaitable<void> {
     $autorun_status = await Control::checkScriptRunning('autorun');
     if ($autorun_status === false) {
-      $document_root = must_have_string(Utils::getSERVER(), 'DOCUMENT_ROOT');
+      $autorun_location = escapeshellarg(
+        must_have_string(Utils::getSERVER(), 'DOCUMENT_ROOT').
+        '/scripts/autorun.php',
+      );
       $cmd =
         'hhvm -vRepo.Central.Path=/var/run/hhvm/.hhvm.hhbc_autorun '.
-        $document_root.
-        '/scripts/autorun.php > /dev/null 2>&1 & echo $!';
+        $autorun_location.
+        ' > /dev/null 2>&1 & echo $!';
       $pid = shell_exec($cmd);
       await Control::genStartScriptLog(intval($pid), 'autorun', $cmd);
     }
--- a/src/models/Control.php
+++ b/src/models/Control.php
@@ -216,6 +216,92 @@ class Control extends Model {
     await Level::genBaseScoring();
   }
 
+  public static async function genAutoBegin(): Awaitable<void> {
+    // Get start time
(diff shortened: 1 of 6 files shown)


justinwray:
This resolves issue #10.

javuto:
This is really cool, and opens the door to have hints or challenges open at a given time! Could we make the 30 seconds to be configurable from the admin page? Same as we do with the bases interval and the progressive scoreboard.

justinwray:
Thanks @javuto!

It should be fairly easy to add additional scheduled tasks into Control::genAutoRun(), so adding time-locked Levels/Hints would be very cool.

Allowing the cycle time to be configurable is a great idea!  I've updated the PR so that the cycle time is now configurable (from the admin configuration page).

src/models/Control.php
```diff
@@ -216,6 +216,92 @@ class Control extends Model {
...
+    }
+  }
+
+  public static async function genRunAutoRunScript(): Awaitable<void> {
+    $autorun_status = await Control::checkScriptRunning('autorun');
+    if ($autorun_status === false) {
+      $document_root = must_have_string(Utils::getSERVER(), 'DOCUMENT_ROOT');
+      $cmd =
+        'hhvm -vRepo.Central.Path=/var/run/hhvm/.hhvm.hhbc_autorun '.
+        $document_root.
```
gsingh93: @iangcarroll pointed out to me that if the user does a manual install and puts the document root in a non-standard location, it's possible that the document root could contain spaces and mess up this command. It's also possible (although not likely) that the document root could contain unsafe special characters.

gsingh93:
Looks good overall, needs a rebase though for the tests.

src/models/Control.php
justinwray: Good point.  Updated to wrap the script path and filename in `escapeshellarg()`.  This should ensure that spaces are handled correctly.

justinwray:
Updated based on feedback and Rebased.

PR #451, opened 2017-02-12
Attachments and Links Import/Export, Database Restore, and Control Cleanup

justinwray:
* Attachments can now be exported and imported.  On export, attachments are downloaded into a Tar Gzip and securely extracted on import.

* Links and Attachments data is now provided within the Levels export.  Users must import both the Level data and the Attachment files to restore the levels with attachments.

* A database restore option has been added which utilizes the backed up database content.  This overwrites all data in the database.

* The Control page has been reorganized to align the various functionality better.

* Memcached flushing has been added to all relevant data imports.

* Error handling has been added to the various import functions.

### DIFF
--- a/src/models/Attachment.php
+++ b/src/models/Attachment.php
@@ -36,10 +36,6 @@ class Attachment extends Model {
     return $this->levelId;
   }
 
-  public static function getAttachmentDir(): string {
-    return strval(self::attachmentsDir);
-  }
-
   // Create attachment for a given level.
   public static async function genCreate(
     string $file_param,
--- a/src/models/Attachment.php
+++ b/src/models/Attachment.php
@@ -17,6 +17,7 @@ class Attachment extends Model {
     private int $id,
     private int $levelId,
     private string $filename,
+    private string $type,
   ) {}
 
   public function getId(): int {
@@ -27,8 +28,16 @@ class Attachment extends Model {
     return $this->filename;
   }
 
+  public function getType(): string {
+    return $this->type;
+  }
+
   public function getLevelId(): int {
     return $this->levelId;
+  }
+
+  public static function getAttachmentDir(): string {
+    return strval(self::attachmentsDir);
   }
 
   // Create attachment for a given level.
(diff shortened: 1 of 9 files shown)


src/models/Level.php
```diff
@@ -186,6 +203,19 @@ private static function levelFromRow(Map<string, string> $row): Level {
...
+      $links = await Link::genAllLinks($level->getId());
+      $link_array = array();
+      foreach ($links as $link) {
+        $link_array[] = $link->getLink();
+      }
+      $attachments = await Attachment::genAllAttachments($level->getId());
+      $attachment_array = array();
+      foreach ($attachments as $attachment) {
+        $attachment_array[] = [
+          "filename" => $attachment->getFilename(),
```
gsingh93: nit: single quotes for strings

justinwray: Fixed.

src/models/Attachment.php
```diff
@@ -27,10 +28,18 @@ public function getFilename(): string {
...
+  public function getType(): string {
+    return $this->type;
+  }
+
   public function getLevelId(): int {
     return $this->levelId;
   }
 
+  public static function getAttachmentDir(): string {
+    return strval(self::attachmentsDir);
```
justinwray: Switched out the getter usage for the constant.

justinwray:
Updated and Rebased.

Maybe we should look into building our own secure `exec` method, possible with filtering and usage of `escapeshellarg`/`escapeshellcmd`? We could then replace all of the current `exec` usage within the project.

The good news is we are not utilizing user-input in the `exec` command calls.

gsingh93:
> The good news is we are not utilizing user-input in the exec command calls.

Yup, but as long as we don't have it, there's a possibility of making a mistake in the future. As well as the mistake of forgetting to escape things like directories with spaces in them. Speaking of which, I forgot to mention to you to escape document root in this PR, just like the other one. Can you do that in another PR (also escape the attachments directory).

stevcoll:
Patch to resolve the issue with imports failing due to missing attachment/link keys. This is for backwards compatibility with older level exports.

[Level.patch.txt](https://github.com/facebook/fbctf/files/823864/Level.patch.txt)

BitSledge:
How do I use that patch? Just replace Level.php with Level.patch.txt or add the patch to Level.php somewhere?

stevcoll:
See section 2 on [this site](http://www.thegeekstuff.com/2014/12/patch-command-examples/). It's a standard patch file.

Issue #435, opened 2017-02-01
Unable to access fbctf(10.10.10.5)


kamikaze123:
After running "vagrant up," I went to 10.10.10.5. However I got blank page(http response is 500 hphp invoke). 

Is latest fbctf working correctly?

david-sykora:
No, it isn't. For me worked reseting to https://github.com/facebook/fbctf/commit/5e3679848c6cbeded68d65e0d37352a33ef603f2

javuto:
There was an issue with `src/controllers/IndexController.php` but it has been fixed now. Just get the latest version and it will work.

gsingh93:
Travis CI should have caught this regression before the PR was merged, but it didn't and that worries me. @justinwray, can you take a look to see how this got through? I think https://travis-ci.org/facebook/fbctf/builds/195824430 is what should have failed. I think we might just be missing a call to `hh_client` in our test script.

PR #452, opened 2017-02-12
Error Checking During Build Tests

justinwray:
* Execute hh_client during build tests.

* Currently the PHP built-in getimagesizefromstring function is not in the HHVM upstream hhi files, and therefore generates an error.  In the future, once getimagesizefromstring is added upstream, we should use the hh_client exit status.


justinwray:
#450 will need to be merged before this PR, or the build tests will fail.

justinwray:
This resolves the outstanding issue from #435.

justinwray:
Note, the build is failing due to the errors resolved in #450.

extra/run_tests.sh
```diff
@@ -29,3 +29,11 @@ hhvm vendor/phpunit/phpunit/phpunit tests
 echo "[+] Deleting test database"
 mysql -u "$DB_USER" --password="$DB_PWD" -e "DROP DATABASE IF EXISTS $DB;"
 mysql -u "$DB_USER" --password="$DB_PWD" -e "FLUSH PRIVILEGES;"
+
+# In the future, we should use the hh_client exit status.
+# Current there are some PHP built-ins not found in the hhi files upstream in HHVM.
+echo "[+] Verifying HHVM Strict Compliance and Error Checking"
```
gsingh93: Heh, nice hack :P

Can you list which PHP builtins it was complaining about for reference?

justinwray: Thanks! :)

Looks like right now the only one we're using is:  getimagesizefromstring().

PR #464, opened 2017-02-22
Update schema.sql

SriHarshaGajavalli:


iangcarroll:
This is not valid SQL; you need to specify what the foreign key references, and you also need to add commas above the line you're adding.

javuto:
Not only the SQL is invalid, as @iangcarroll said, but the change makes no sense. The links are related to levels, there is no point adding teams here.

PR #494, opened 2017-04-15
Updated LiveSync Security

justinwray:
* Live Sync API is now disabled by default.

* Admins can enable or disable the Live Sync API from the Administration Configuration page.

* Live Sync API now has an optional 'Auth Key.'  When the auth key is set, anyone attempting to pull from the API must supply the auth key value in their request.  Without the auth key, no data is provided by the Live Sync API endpoint.

* When using the Auth Key, it must be added as a parameter to the URL value in the `liveimport` script: ```?auth=XXXXX```

  * Example (with an auth key of `1234567890`:

  * `hhvm -vRepo.Central.Path=/var/run/hhvm/.hhvm.hhbc_liveimport /var/www/fbctf/src/scripts/liveimport.php --url 'https://10.10.10.101/data/livesync.php?auth=1234567890'`

  * Note:  When using the Auth Key you should use a secure key.

* The `livesync` API endpoint will provide error messages if the API is disabled, the key is missing or invalid, or if any general error is encountered.

* The `liveimport` script will check for errors and display those in the output if any are encountered.

### DIFF
--- a/database/schema.sql
+++ b/database/schema.sql
@@ -229,6 +229,8 @@ INSERT INTO `configuration` (field, value, description) VALUES("password_type", 
 INSERT INTO `configuration` (field, value, description) VALUES("default_bonus", "30", "(Integer) Default value for bonus in levels");
 INSERT INTO `configuration` (field, value, description) VALUES("default_bonusdec", "10", "(Integer) Default bonus decrement in levels");
 INSERT INTO `configuration` (field, value, description) VALUES("language", "en", "(String) Language of the system");
+INSERT INTO `configuration` (field, value, description) VALUES("livesync", "0", "(Boolean) LiveSync functionality");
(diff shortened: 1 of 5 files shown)


iangcarroll:
I'd suggest using `hash_equals()` for comparing keys to prevent any timing attacks.

gsingh93:
ping @justinwray

justinwray:
@gsingh93 Not a problem at all.

We have already implemented and tested @iangcarroll's recommendation.

We are currently working through the performance implications of the Awaitables. There are a number of loops that have internal awaits and we want to ensure we combine in a way that doesn't cause performance degradation from a redundant loop perspective.

This will defiantly be completed/updated this week.

justinwray:
Updated based on feedback.

Issue #500, opened 2017-05-12
fbctf plattform doesn't start anymore - modules missing


PhilippBehmer:
Using the latest master (e880251b4ed71c499d7a97673c196311a40e79a3)
Fresh & updated Ubuntu 14.04 AWS instance: Linux ip-172-31-28-207 3.13.0-116-generic #163-Ubuntu SMP Fri Mar 31 14:13:22 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
Manuall provisioning using the command: ./extra/provision.sh -m prod -s $PWD

Caused by the "grunt" command in the "run_grunt()" function in the "lib.sh" file.

Log output:

```
[...]
/usr/bin/flow -> /usr/lib/node_modules/flow-bin/cli.js
flow-bin@0.46.0 /usr/lib/node_modules/flow-bin
Running "force:eslint" (force) task

Running "force-internal:pre:eslint" (force-internal) task
>> Enable force mode for task eslint.

Running "eslint:dist" (eslint) task

/var/www/fbctf/node_modules/eslint-plugin-flowtype/dist/rules/defineFlowType.js:37
    ClassImplements(node) {
                   ^
Warning: Failed to load plugin flowtype: Unexpected token ( Used --force, continuing.
Warning: Task "eslint:dist" failed. Used --force, continuing.

Running "force-internal:post:eslint" (force-internal) task
>> Reset force mode for task eslint.

Running "run:flow" (run) task
flow is still initializing; this can take some time. [processing] -Launching Flow server for /var/www/fbctf/src
Spawned flow server (pid=11427)
Logs will go to /tmp/flow/zSvarzSwwwzSfbctfzSsrc.log
Started a new flow server: \flow is still initializing; this can take some time. [local inference] |src/static/js/admin.js:5
  5: var Dropkick = require('dropkickjs');
                            ^^^^^^^^^^^^ dropkickjs. Required module not found

src/static/js/admin.js:6
  6: var $ = require('jquery');
                     ^^^^^^^^ jquery. Required module not found

src/static/js/clock.js:3
  3: var $ = require('jquery');
                     ^^^^^^^^ jquery. Required module not found

src/static/js/fb-ctf.js:12
 12: var d3 = require('d3');
                      ^^^^ d3. Required module not found

src/static/js/fb-ctf.js:13
 13: var $ = require('jquery');
                     ^^^^^^^^ jquery. Required module not found

src/static/js/fb-ctf.js:14
[...]

Issue #499, opened 2017-05-10
docker dev environment failing to build


m0jtaba:
Cloned master

command: docker build --build-arg MODE=dev -t="fbctf_in_dev" .

Only change I have made is in: /fbctf/extra/provision.sh: which is: **set -e -x**

anyone have any ideas?

Output:

`+ sudo npm install -g flow-bin
/usr/bin/flow -> /usr/lib/node_modules/flow-bin/cli.js
flow-bin@0.46.0 /usr/lib/node_modules/flow-bin
+ run_grunt /var/www/fbctf dev
+ local __path=/var/www/fbctf
+ local __mode=dev
+ cd /var/www/fbctf
+ grunt
Running "force:eslint" (force) task

Running "force-internal:pre:eslint" (force-internal) task
>> Enable force mode for task eslint.

Running "eslint:dist" (eslint) task

/var/www/fbctf/node_modules/eslint-plugin-flowtype/dist/rules/defineFlowType.js:37
    ClassImplements(node) {
                   ^
Warning: Failed to load plugin flowtype: Unexpected token ( Used --force, continuing.
Warning: Task "eslint:dist" failed. Used --force, continuing.

Running "force-internal:post:eslint" (force-internal) task
>> Reset force mode for task eslint.

Running "run:flow" (run) task
flow is still initializing; this can take some time. [processing] -Launching Flow server for /var/www/fbctf/src
Spawned flow server (pid=9037)
Logs will go to /tmp/flow/zSvarzSwwwzSfbctfzSsrc.log
Started a new flow server: \flow is still initializing; this can take some time. [merging inference] |src/static/js/admin.js:5
  5: var Dropkick = require('dropkickjs');
                            ^^^^^^^^^^^^ dropkickjs. Required module not found

src/static/js/admin.js:6
  6: var $ = require('jquery');
                     ^^^^^^^^ jquery. Required module not found

src/static/js/clock.js:3
  3: var $ = require('jquery');
                     ^^^^^^^^ jquery. Required module not found

src/static/js/fb-ctf.js:12
 12: var d3 = require('d3');
                      ^^^^ d3. Required module not found

src/static/js/fb-ctf.js:13
 13: var $ = require('jquery');
                     ^^^^^^^^ jquery. Required module not found

src/static/js/fb-ctf.js:14
 14: var Keycode = require('keycode');
                           ^^^^^^^^^ keycode. Required module not found

src/static/js/filter.js:3
  3: var $ = require('jquery');
[...]

PR #502, opened 2017-05-13
Temp fix for issue 499 & 500 - Force Grunt to continue

wisco24:
Forcing Grunt to continue as it is not correctly detecting node_modules in the folder.  

#499 & #500

### DIFF
--- a/extra/lib.sh
+++ b/extra/lib.sh
@@ -80,7 +80,7 @@ function run_grunt() {
   local __mode=$2
 
   cd "$__path"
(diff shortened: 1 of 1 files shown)


stevcoll:
This fix certainly works, but I'm wondering if there are any larger or longer-term implications of using grunt --force. Running this by the project leads to see if this solution will work, or we need to come up a more specific way to address this problem.

gsingh93:
Merged.

PR #512, opened 2017-05-27
Level Deletion Confirmation and Bug Fixes

justinwray:
* Added deletion confirmation dialog when deleting levels.

* Level data is now deleted from HintLog, ScoresLog, and FailuresLog, when a level is deleted.

* Cache invalidation performed on level deletion for records containing levels.

### DIFF
--- a/src/models/Level.php
+++ b/src/models/Level.php
@@ -627,13 +627,16 @@ class Level extends Model implements Importable, Exportable {
 
     // Remove team points for level
     $scores = await ScoreLog::genAllScoresByLevel($level_id);
+    $level_delete_queries = Vector {};
     foreach ($scores as $score) {
       $team_id = $score->getTeamId();
       $points = $score->getPoints();
-      await $db->queryf(
-        'UPDATE teams SET points = points - %d WHERE id = %d',
-        $points,
-        $team_id,
+      $level_delete_queries->add(
+        sprintf(
+          'UPDATE teams SET points = points - %d WHERE id = %d',
+          $points,
+          $team_id,
+        ),
       );
     }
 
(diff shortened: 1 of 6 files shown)


justinwray:
This commit includes the changes found in PR #509. Once #509 has been merged, this PR can be rebased.

justinwray:
This PR is now mergeable.

Given rebase conflicts: the branch was reset (hard) to the upstream dev branch, and then the PR changes were patched and merged onto dev, submitted via a (force) push. All relevant commit log data has been retained in the final commit message.

src/models/Level.php
```diff
@@ -625,9 +625,47 @@ private static function levelFromRow(Map<string, string> $row): Level {
...
+      await $db->queryf(
+        'UPDATE teams SET points = points + %d WHERE id = %d',
+        $penalty,
+        $team_id,
+      );
+    }
+
+    // Delete all references to level
     await $db->queryf('DELETE FROM levels WHERE id = %d LIMIT 1', $level_id);
+    await $db->queryf('DELETE FROM hints_log WHERE level_id = %d', $level_id);
```
gsingh93: Remember to try and combine awaits together with something like `\HH\Asio\va`

You can do that for loops too by putting the `Awaitables` in an array.

justinwray: Moved all awaitable queries into a vector which are now executed via AsyncMysqlConnection::multiQuery().

justinwray:
Updated based on feedback.

gsingh93:
The use of `sprintf` here is fine because we're replacing `id`, but just remember in the general case it's not fine and will lead to SQL injection. When dealing with multiple unsafe queries we need to collect all the Awaitables from `queryf` in an array, and then use `\HH\Asio\v` or a similar function to convert the array of Awaitables into a single Awaitable, and then Await on that.

But since these queries only use `id`, which is not controlled by the user, I'll go ahead an accept this.

PR #537, opened 2017-07-24
Fix #421 - Update to have hint penalty only assessed after scoring th…

### LINKED ISSUE #421: Hints
How can i do that hints substract the points only when the flag its ok?

Sorry for my english

ArchAng31:
Moved the assessing of penalty for a hint from county-data.php to Level.php so the penalty is now assessed when the level is scored and not when the hint button is first pressed.


justinwray:
This PR should be against the `dev` branch.  It will also need to be rebased, which should resolve the build error.

ArchAng31:
rebasing against dev.

Issue #218, opened 2016-09-20
Multiple Choice Quiz Questions


wisco24:
Is there any way to also make Quiz questions into multiple choice instead of just fill-in-the-blank?  

If not, I think this would be a good Feature Request.

Issue #19, opened 2016-05-11
Allow setting bonus, bonus decrement, attachments, and links when creating a challenge


gsingh93:
We currently require you to create a challenge first before setting any of these values (for legacy reasons), but now there's no reason not to be able to add these when we create a challenge.

PR #540, opened 2017-08-03
Added multiple choice #218; can set bonuses initially #19

ArchAng31:
Added functionality for multiple choice choice questions when making a Quiz. Also, updated functionality to set bonus and bonus decr when initially setting the problem. Finally, added functionality for a "wrong answer penalty" to quizes and flags (required to prevent multiple choice spamming).


justinwray:
This PR should be against the `dev` branch.  It will also need to be rebased, which should resolve the build error.

ArchAng31:
Rebasing against dev.

Issue #550, opened 2017-08-24
NPM error when try to build dev environment


ayr-ton:
HEAD version 3047351e81cc8e33b616c0db2aa24c2e1d584e50

It is happening both with the manual, vagrant and docker builds.

Following the default steps on README, using Mac OS Sierra as the development machine, I got the following NPM error:
```
(supressed the beggining of the input, as everything working as expected so far)
[+] Installing ca-certificates
Reading package lists... Done
Building dependency tree
Reading state information... Done
ca-certificates is already the newest version.
0 upgraded, 0 newly installed, 0 to remove and 3 not upgraded.
[+] Installing npm
Reading package lists... Done
Building dependency tree
Reading state information... Done
npm is already the newest version.
0 upgraded, 0 newly installed, 0 to remove and 3 not upgraded.
[+] Updating npm

/usr/local/lib/node_modules/npm/lib/config/cmd-list.js:113
module.exports.aliases = Object.assign({}, shorthands, affordances)
                                ^
TypeError: Object function Object() { [native code] } has no method 'assign'
    at Object.<anonymous> (/usr/local/lib/node_modules/npm/lib/config/cmd-list.js:113:33)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at /usr/local/lib/node_modules/npm/lib/npm.js:37:17
    at Object.<anonymous> (/usr/local/lib/node_modules/npm/lib/npm.js:471:3)
    at Module._compile (module.js:456:26)
```

wisco24:
I'm having the same issue on Ubuntu 14 with the current code, and previous code.

ayr-ton:
@wisco24 Do you remember the last build that was working?

wisco24:
I think it has to do with a new version of a dependency.

stevcoll:
There is a PR almost complete, just doing some final testing.

ayr-ton:
@stevcoll If you want us to test the PR, let us know \o/

juanrossi:
@stevcoll same as Ayrton, let me know if I can help with anything or testing, I need to run a deploy tonight so this is going to help a lot! thanks.

skcollins:
I banged my head on this one for a bit. Seems like a dependency issue in the provision.sh script. I commented out 
```
#package npm
[...]

PR #551, opened 2017-08-25
Remove NodeJS-Legacy (Fixes: #550)

justinwray:
* Updated version of node.js installed to fix critical provisioning errors. By default, Ubuntu 14.04 uses a legacy version of node.js. This code removes the legacy version and updates to a newer version.

* Source: https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions

* This PR fixes the provision/build issue from #550


ayr-ton:
+1

ayr-ton:
It fixes #550

PR #555, opened 2017-08-28
Downgraded DropkickJS and Streamlined Provision for NodeJS/Downloads (Fixes: #554)

### LINKED ISSUE #554: Provision Failing - Dropkickjs SyntaxError: 'import' and 'export' may only appear at the top level
The following error is currently printed during provisioning:

```
Running "browserify:dist" (browserify) task
>> SyntaxError: 'import' and 'export' may only appear at the top level (2:0) while parsing /var/www/fbctf/node_modules/dropkickjs/src/dropkick.js while parsing file: /var/www/fbctf/node_modules/dropkickjs/src/dropkick.js
```
This causes the entire copy:browserify task to fail. Provision completes, but the entire dist directory is missing, and thus the platform is unusable.

justinwray:
* Downgraded Dropkick.js to version 2.1.10.  The project originally was built using 2.1.10 and specified a near version in the 2.x.x release family.  On August 27th 2017 Dropkick.js released version 2.2.0 which is incompatible with ES6 specs. The incompatibility with the new release of Dropkick.js caused the provisioning of the platform to fail.

* Moved the installation process for Node.js to a function within `lib.sh`.  This change streamlines the provision script.

* Removed the installation of `wget` from provisioning.  `wget` is no longer used within the project and is therefore unneeded.

* Updated the `dl()` download function within the provision script to use `curl` exclusively, with retry options.  The retry options are set to 5 retries with a 15-second delay between retries.  The addition of the retry option ensures the provision can continue if there is a temporary issue with a remote connection or availability of a remote resource.

* Added the `dl_pipe()` download function to the provision script.  This download function provided the data from the remote resource via standard output to be piped into another command.  As piping downloads within the provisioning process have become more common, this function streamlines the process.

* Fixes #554 

* Updates fixes for #550

### DIFF
--- a/extra/lib.sh
+++ b/extra/lib.sh
@@ -29,12 +29,12 @@ function ok_log() {
 function dl() {
   local __url=$1
   local __dest=$2
+  sudo curl --retry 5 --retry-delay 15 -sSL "$__url" -o "$__dest"
+}
 
-  if [ -n "$(which wget)" ]; then
-    sudo wget -q "$__url" -O "$__dest"
-  else
-    sudo curl -s "$__url" -o "$__dest"
-  fi
+function dl_pipe() {
+  local __url=$1
+  curl --retry 5 --retry-delay 15 -sSL "$__url"
 }
 
 function package_repo_update() {
@@ -53,7 +53,7 @@ function package() {
 
 function install_unison() {
   cd /
-  curl -sL https://www.archlinux.org/packages/extra/x86_64/unison/download/ | sudo tar Jx
+  dl_pipe "https://www.archlinux.org/packages/extra/x86_64/unison/download/" | sudo tar Jx
 }
 
 function repo_osquery() {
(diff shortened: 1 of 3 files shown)


dparker18:
I'm seeing some new errors trying to build prod environment with docker

First couple errors follow, full log attached:

Running "run:flow" (run) task
flow is still initializing; this can take some time. [processing] -Launching Flow server for /var/www/fbctf/src
Spawned flow server (pid=5264)
Logs will go to /tmp/flow/zSvarzSwwwzSfbctfzSsrc.log
Started a new flow server: \flow is still initializing; this can take some time. [processing] |Error: src/static/js/admin.js:5
  5: var Dropkick = require('dropkickjs');
                            ^^^^^^^^^^^^ dropkickjs. Required module not found

Error: src/static/js/admin.js:6
  6: var $ = require('jquery');
                     ^^^^^^^^ jquery. Required module not found

Any ideas? I'll try dev instead of prod and report if that's any different, thank you!

[prodbuild.log.txt](https://github.com/facebook/fbctf/files/1255270/prodbuild.log.txt)

justinwray:
@dparker18 

Those errors stem from issues #499 and #500, which were addressed in PR #502.  Grunt is (_safely_, in this case) ignoring those errors and continuing the build process.

So even though you see those errors in the log, it should still be successfully building.

That said, your log clearly shows that the build is failing.  From a quick glance the failure seems to be with the `certbot` SSL process:

```
/root/extra/lib.sh: line 131: /root/tmp/certbot.sh: No such file or directory
```

Please open a new Issue (with relevant information), and we can track and troubleshoot that specific issue.

dparker18:
Good catch. Looks like a simple "mkdir -p /root/tmp" in extra/lib.sh fixed it, will verify and PR tomorrow

justinwray:
@ethanaldrich9986 

This PR, along with #551, fix a few Node related errors.  If you are still seeing errors after these updates are pushed to the `master` branch please open a new Issue (with relevant information), and we can track and troubleshoot that specific issue.

Issue #548, opened 2017-08-22
Inefficient MySQL Session Queries


juanrossi:
Hello, I've doing some load testing to the last version of CTF (https://github.com/facebook/fbctf/commit/ed0a225708ba83cb7a7e7b818bede9ed37da1bad) and saw that some queries and the DB in general could be optimised.

The `session` table has the column `cookie` with type `text`. This is inefficient because we have a fixed size value that could fit inside a `VARCHAR(200)` without any issue. Also, the text type doesn't save values to memory, which generates lot of unnecessary CPU usage.

I know that on a normal usage, all this information would be also saved to Memcached, but I don't see why we can't also optimise this.

![screen shot 2017-08-22 at 2 27 09 pm](https://user-images.githubusercontent.com/569417/29578606-2db9fa76-8746-11e7-99be-e05e0be05f5d.png)
This graph represents the CPU usage (With 100+ concurrent requests) after changing the type to `varchar` and adding an INDEX to `cookie`.

Also, I haven't seen any index besides the primary key ones or a UNIQUE key.

Would you be interested in a PR with improvements on the database schema?

stevcoll:
@juanrossi Thanks for for research! We would certainly appreciate any optimizations you could suggest for the database schema if possible. There are some other priorities right now so we haven't had a lot of time to look into these areas.

wisco24:
Linking Issue #456

stevcoll:
Closing as there is a PR #564 for this that should be merged soon.

PR #564, opened 2017-09-03
Optimization of connections and database

juanrossi:
Hello, this pull request should improve the database schema and making correct use of database types (For example session and use of INDEX [#548]).

In HHVM the pool of MySQL connections is created by request, which is now opening between 5 and 15 connections depending on the section. This generated a big overhead to the database, which will result in max_allowed_connections errors with 50 concurrent users.

Changing the idle timeout of pool from 0.2 to 2 seconds will result in reutilization of connections instead of generating new ones and 20 max connections to make sure the db is not stressed (Default is 50).

Let me know if have any questions!

### DIFF
--- a/extra/hhvm.conf
+++ b/extra/hhvm.conf
@@ -24,3 +24,12 @@ hhvm.repo.central.path = /var/run/hhvm/hhvm.hhbc
 hhvm.mysql.socket = /var/run/mysqld/mysqld.sock
 hhvm.pdo_mysql.socket = /var/run/mysqld/mysqld.sock
 hhvm.mysqli.socket = /var/run/mysqld/mysqld.sock
+
+# MySQL Optimization
+# Use persistent connections and higher timeout
+mysqli.reconnect = 1
+mysqli.allow_persistent = On
+mysqli.max_persistent = 400
+hhvm.mysql.connect_timeout = 1500
+hhvm.mysql.read_timeout = 2000
+hhvm.mysql.max_retry_open_on_fail = 3
--- a/src/Db.php
+++ b/src/Db.php
@@ -17,8 +17,10 @@ class Db {
   private function __construct() {
     $this->config = parse_ini_file($this->settings_file);
     $options = array(
-      'idle_timeout_micros' => 200000,
+      'idle_timeout_micros' => 2000000,
       'expiration_policy' => 'IdleTime',
+      'per_key_connection_limit' => 20,
+      'pool_connection_limit' => 20
     );
     $this->pool = new AsyncMysqlConnectionPool($options);
   }
--- a/database/countries.sql
+++ b/database/countries.sql
@@ -12,7 +12,9 @@ CREATE TABLE `countries` (
   `enabled` tinyint(1) DEFAULT 0,
   `d` text DEFAULT NULL,
   `transform` text DEFAULT NULL,
-  PRIMARY KEY (`id`)
+  PRIMARY KEY (`id`),
+  KEY `iso_code` (`iso_code`),
+  KEY `enabled` (`enabled`)
 ) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
--- a/database/schema.sql
+++ b/database/schema.sql
@@ -36,7 +36,7 @@ CREATE TABLE `levels` (
   `id` int(11) NOT NULL AUTO_INCREMENT,
   `active` tinyint(1) NOT NULL,
   `type` varchar(4) NOT NULL,
-  `title` text NOT NULL,
+  `title` varchar(255) NOT NULL,
   `description` text NOT NULL,
   `entity_id` int(11) NOT NULL,
   `category_id` int(11) NOT NULL,
@@ -48,7 +48,9 @@ CREATE TABLE `levels` (
   `hint` text NOT NULL,
   `penalty` int(11) NOT NULL,
   `created_ts` timestamp NOT NULL DEFAULT 0,
-  PRIMARY KEY (`id`)
+  PRIMARY KEY (`id`),
+  KEY `entity_id` (`entity_id`),
+  KEY `active` (`active`)
 ) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
@@ -61,7 +63,7 @@ DROP TABLE IF EXISTS `categories`;
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE `categories` (
   `id` int(11) NOT NULL AUTO_INCREMENT,
-  `category` text NOT NULL,
+  `category` varchar(255) NOT NULL,
   `protected` tinyint(1) NOT NULL,
   `created_ts` timestamp NOT NULL DEFAULT 0,
   PRIMARY KEY (`id`)
@@ -116,8 +118,8 @@ DROP TABLE IF EXISTS `teams`;
 CREATE TABLE `teams` (
   `id` int(11) NOT NULL AUTO_INCREMENT,
   `active` tinyint(1) NOT NULL DEFAULT 1,
-  `name` text NOT NULL,
-  `password_hash` text NOT NULL,
+  `name` varchar(255) NOT NULL,
+  `password_hash` varchar(255) NOT NULL,
   `points` int(11) NOT NULL DEFAULT 0,
   `last_score` timestamp NOT NULL,
   `logo` text NOT NULL,
@@ -125,7 +127,9 @@ CREATE TABLE `teams` (
   `protected` tinyint(1) NOT NULL DEFAULT 0,
   `visible` tinyint(1) NOT NULL DEFAULT 1,
   `created_ts` timestamp NOT NULL DEFAULT 0,
-  PRIMARY KEY (`id`)
+  PRIMARY KEY (`id`),
+  KEY `visible` (`visible`),
+  KEY `active` (`active`)
 ) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
@@ -157,9 +161,10 @@ CREATE TABLE `teams_data` (
   `id` int(11) NOT NULL AUTO_INCREMENT,
   `team_id` int(11) NOT NULL,
   `name` text NOT NULL,
-  `email` text NOT NULL,
+  `email` varchar(255) NOT NULL,
   `created_ts` timestamp NOT NULL DEFAULT 0,
-  PRIMARY KEY (`id`)
+  PRIMARY KEY (`id`),
+  KEY `team_id` (`team_id`)
 ) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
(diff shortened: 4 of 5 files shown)


justinwray:
@juanrossi Thank you for the submission!  We will review and test the database optimization changes submitted here and then provide any feedback we might have.

stevcoll:
@juanrossi Great work! I noticed a significant speed increase while using the platform. Testing looks good. @justinwray ready for your review.

wisco24:
How long is it going to take to get this merged in??!?!?!?  This issue was identified a long time ago but the admin's didn't believe it.   Please test and approve this ASAP.

justinwray:
@juanrossi 

We have had a chance to test and review your submission.  The modifications to the schema are great.  Specifically, the indexing provides a valuable optimization with very little overhead.

The one area we are concerned about within this PR is the inclusion of persistent connections.  While on the surface persistent connections sound like a performance improvement they bring along a large number of downfalls.  In fact, often the use of persistent connections can negatively impact performance.  These issues are well researched, so take a look at:  http://php.net/manual/en/features.persistent-connections.php and https://meta.wikimedia.org/wiki/Why_persistent_connections_are_bad.

Generally, one would utilize persistent connections when the TCP/IP overhead of the database connections are increasing latency.  Given that the most common deployment of the platform has the MySQL server on the very same system, the connections, in this case, are all contained locally.  In other words, there is almost no TCP/IP overhead in the most common deployment.  Even still, if one were to deploy the database to a different server (using the new installation option, or via a cloud solution like RDS), the distance from the platform stack and the database should be relatively small.  In the extreme cases where a deployment is setup with a database that is significantly far (network path wise) from the rest of the stack, persistent connections _may_ provide some benefit.

Thankfully, as you outlined in your Issue (#548), we are caching the database responses in Memcached.  Even sessions queries are cached.  So the number of connections to MySQL directly utilized is fairly minimal.

As such, we believe it would be better to leave the inclusion of persistent connections left up to the end-user.  We believe persistent connections to be an uncommon option and in most cases to have negative consequences.

Could you update this PR with the removal of the persistent connection change, or make it optional (through the provisioning process)?

fredemmott:
Is this actually enabling persistent connections? I see the option is being set //for mysqli//, but according to github search, this project isn't using mysqli (though options are also set in a few other places), and the mysqli options shouldn't affect the async mysql extension

juanrossi:
Hi @justinwray, thanks for the review and information.

Have in mind that right now the framework (CTF) and HHVM is generating several connections per request (Between 5 to 15 for EACH request) which is a lot. With sessions and memcached I saw between 4 to 7 new connections being created in the game (around 20 in total, but 15 being reused). This generates a lot of overhead.

So, we can stop using async connections and make a change to Db.php:genConnection to something like this (Without the Async):

```
public async function genConnection(): Awaitable<AsyncMysqlConnection> {
    if(!$this->isConnected()) {
        await $this->genConnect();
        invariant($this->conn !== null, 'Connection cant be null.');
        return $this->conn;
    }
    return $this->conn;
}
```

This will create one connection and re-use (with Async I saw a few errors, so this is not going to work directly with that change).

The other option is to embrace the persistent connections (which is not bad, you just have to use them correctly), and reuse them when possible.

You can see in [HHVM MySQL documentation](https://docs.hhvm.com/hack/reference/class/AsyncMysqlConnectionPool/) that we are creating a pool of connections for each request. And you can also validate this in the game adding `var_dump($this->pool->getPoolStats());` after setting `$this->conn`.

Let me know if you have questions.

fredemmott:
Memoizing async connections like that is definitely unsafe, given that an async function can make another mysql request while another is waiting for a response.

juanrossi:
@fredemmott That's correct, that's why we shouldn't use that code with Async, it was just as an example.

I think we should use async (with persistent connections which HHVM uses), or change ALL the queries to use synchronous connections. I'd go for the first option.

fredemmott:
Yeah, I'm fine with us using async mysql's pool, that's pretty much a requirement (but we need to be sure there's nothing setting options on the connection or otherwise mutating state)

This is what confuses me:

```
+# Use persistent connections and higher timeout
 +mysqli.reconnect = 1
 +mysqli.allow_persistent = On
 +mysqli.max_persistent = 400
```

What are those doing on FBCTF, given FBCTF doesn't use mysqli?

justinwray:
> @fredemmott 
> Is this actually enabling persistent connections? I see the option is being set //for mysqli//, but according to github search, this project isn't using mysqli (though options are also set in a few other places), and the mysqli options shouldn't affect the async mysql extension

This is a good point.  The change is impacting `mysqli`, not `AsyncMysqlConnectionPool`.

> @juanrossi 
>This will create one connection and re-use (with Async I saw a few errors, so this is not going to work directly with that change).

Please do not misunderstand, I have nothing against persistent connections.  When used properly they can be valuable.  (MediaWiki used the term _'bad'_)

> @fredemmott 
> Memoizing async connections like that is definitely unsafe, given that an async function can make another mysql request while another is waiting for a response.

Agreed.  This sort of re-use introduces a lot of hard-to-track bugs and worse complete lock ups in some cases (or at least such is true in `mysqli`).  Same goes for persistent connections in general, when not throughly planned around.

@juanrossi:

We are using [`AsyncMysqlConnectionPool`](https://docs.hhvm.com/hack/reference/class/AsyncMysqlConnectionPool/) which should be giving us re-use of connections in the pool:

> The connection pool will either create a new connection or use one of the recently available connections from the pool itself.

Much like you would experience with persistent connections in native `mysqli`.

juanrossi:
Hi guys, after reading some [posts](http://hhvm.com/blog/3689/implementing-mysqli), [docs](https://docs.hhvm.com/hack/async/extensions ) and [issues](https://github.com/facebook/hhvm/issues/6734) I thought it was necessary, but I might misunderstood that.

If you are sure it's not required and we are not using MySQLi, I'll remove those lines.

justinwray:
@juanrossi 

Given the very clear benefits of the indexing, and the contention over persistent connections:  Let's remove the persistent connections from this PR and we will merge the schema updates.

A discussion can then be started in [Issues](https://github.com/facebook/fbctf/issues) to further discuss the implementation and implications of persistent connections.

juanrossi:
@justinwray sorry for the delay, I pushed the commit removing MySQLi vars. 

Let me know if you think we need to add more changes to this!

juanrossi:
Hi @justinwray, did you have the chance to review this changes? Thanks.

src/Db.php
```diff
@@ -17,8 +17,10 @@ public static function getInstance(): Db {
   private function __construct() {
     $this->config = parse_ini_file($this->settings_file);
     $options = array(
-      'idle_timeout_micros' => 200000,
+      'idle_timeout_micros' => 2000000,
```
justinwray: > 'idle_timeout_micros' => 2000000,

**Can you explain the justification for the value chosen?**

The default value is `4000000` microseconds or 4 seconds.  The value selected here is `2000000` or 2 seconds, which is lower than the default, though higher than what is currently set within the project.

Given we are using an `expiration_policy` of `IdleTime`, this value will determine when the connections in the pool are destroyed.  While this does not directly have any relation to persistent connections, the implications, in this case, are similar.  Given the amount of caching, database queries should be rarer than they previously were within the platform, and a value of 0.2 seconds (the old value) is likely too low, resulting in an increased volume of connection overhead.  However, 2 seconds might also be too little as well.  It seems likely that the new Cache/DB design would benefit from an increased value here.

Do you have any test data that provides merit to this particular value?

src/Db.php
```diff
@@ -17,8 +17,10 @@ public static function getInstance(): Db {
...
       'expiration_policy' => 'IdleTime',
+      'per_key_connection_limit' => 20,
```
justinwray: > `per_key_connection_limit' => 20`

**Can you explain the justification for lowering this from the default value?**

The default is `50` connections.  Lowering this to `20` means that if any additional connections are requested after 20 exist in the pool, an exception will be thrown and the connection will not be created.  _It should be noted that this value is specific to number of allowed connections per MySQL server hostname, port, database, and username._

In optimal conditions, lowering this value to `20` would not cause any issue.  In fact, when things are running smoothly, the maximum number of connections in the pool is only `2`.  Currently a full AJAX refresh of the gameboard requests just over `50` connections, but the project maintains the default of `setReusable` so connections in the pool that are idle will be reused (or safely recreated if they have been destroyed).

The potential issue arises when those connections in the pool are not idle.  Imagine a scenario where there is a high load on the database and the gameboard AJAX refresh is requesting _just over_ `50` connections, and none of the connections in the pool are available for reuse yet (they are still requesting/retrieving data).

To be honest, given the way the project is structured (especially, with the caching) a connection pool with `50` or even `20` connections seems highly unlikely, and would likely mean other timeouts are being hit or things are failing.  So I do not think there is much risk in lowering this value.  However, there doesn't seem to be much benefit, especially with the ease of increasing the MySQL connection limit (`max_connections`).  Again, I believe erring on the side of slow data over no data is appropriate.

Do you have any test data that provides credence to lowering this particular value?

src/Db.php
```diff
@@ -17,8 +17,10 @@ public static function getInstance(): Db {
...
+      'pool_connection_limit' => 20
```
justinwray: > `pool_connection_limit' => 20`

**Can you explain the justification for lowering this from the default value?**

The default is `50` connections.  _It should be noted that this value is the total number of pool connections, regardless of the MySQL server hostname, port, database, and username._

In theory, this value would be irrelevant, as the `per_key_connection_limit` would be hit when using a single database server.  However, it is possible to use multiple databases.  In fact, PR #419 provides the ability to use database replication across numerous replicated databases (though, to be honest, in its current form #419 is highly unlikely to ever be merged).  Regardless, a user could utilize multiple databases which would bring this value into relevancy.

Please see the comment on the line above (`per_key_connection_limit' => 20`) for more details and thoughts on this change.

extra/hhvm.conf
```diff
@@ -24,3 +24,9 @@ hhvm.repo.central.path = /var/run/hhvm/hhvm.hhbc
 hhvm.mysql.socket = /var/run/mysqld/mysqld.sock
 hhvm.pdo_mysql.socket = /var/run/mysqld/mysqld.sock
 hhvm.mysqli.socket = /var/run/mysqld/mysqld.sock
+
+# MySQL Optimization
+# Use persistent connections and higher timeout
```
justinwray: Nit:  You should remove the no-longer-accurate comment about _persistent connections_.

extra/hhvm.conf
```diff
@@ -24,3 +24,9 @@ hhvm.repo.central.path = /var/run/hhvm/hhvm.hhbc
...
+hhvm.mysql.connect_timeout = 1500
```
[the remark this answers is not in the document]
juanrossi: While testing the platform with heavy load (200+ concurrent users, which is not a lot) I saw a lot of timeouts connecting the database, so this helped a bit.

extra/hhvm.conf
```diff
@@ -24,3 +24,9 @@ hhvm.repo.central.path = /var/run/hhvm/hhvm.hhbc
...
+hhvm.mysql.read_timeout = 2000
```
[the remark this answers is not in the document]
juanrossi: Yeah, I kind of agree on the "it's preferred to return data", but imho I don't think any user will wait more than 5 seconds for a requests (and I think 5 it's already a lot).

extra/hhvm.conf
```diff
@@ -24,3 +24,9 @@ hhvm.repo.central.path = /var/run/hhvm/hhvm.hhbc
...
+hhvm.mysql.max_retry_open_on_fail = 3
```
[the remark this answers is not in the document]
juanrossi: I tested with a heavy load and saw a small improvement with this, but I don't think it's such a big change. I could lower it to two, just to be on the safe side.

src/Db.php
juanrossi: I think with 4 seconds would be ok, but 0.2 was too low and saw that the requests were generating way too many new connections. Also, a high value (like 10s) will keep the connections for a longer time which could also bring some issues if the database doesn't have a lot of available connections in the pool.

I think we should set this to a number we feel confortable but not leave the default of the framework. If this changes tomorrow to a lower value we'll see a lot of performance issues imo.

src/Db.php
juanrossi: I think 50 connections is too much. I agree that requests are going to fail, but I saw that 20 was a number were requests didn't fail and I didn't risk taking all of my available connections (which were around 1500).

I think this is a "hack" and we should improve the reuse of connections, because I was some requests creating up to 17 connections for a SINGLE requests. When you have more than 200 concurrent users playing, that's a LOT. I preferred returning some errors than saturating my MySQL to all my users.

src/Db.php
juanrossi: I answered the other comment, I agree that this is irrelevant if we have the other line.

wisco24:
Any idea when this will be merged?  Getting close to 6 months....

juanrossi:
Hi @wisco24, although it's right it's been a while, most of the optimizations have been implemented in PR #594.

Issue #563, opened 2017-09-03
Errors in game when the session expires


juanrossi:
Hello, I just had an event with more than 200 concurrent players and had a few error 500.

Searching the case I found out that when the session expires and the players are on the gameboard, they are never redirected to the index, and the server continues to receives requests.

When the `enforceLogin()` function is called, and the session is not active I'm having the following error:

```
[Sun Sep  3 14:46:12 2017] [hphp] [23:7f18aa7ff700:63511:000001] [] \nFatal error: Uncaught exception 'IndexRedirectException' with message '' in /var/www/fbctf/src/SessionUtils.php:113\nStack trace:\n#0 /var/www/fbctf/src/data/teams.php(7): SessionUtils::enforceLogin()\n#1 {main}
```

Although this error should not happen, if it didn't we would be getting a 302 and redirecting all request to index, but the requests would still be happening (and be larger ones).

I think the frontend should validate if the session is valid or not every few requests or minutes, and in case it's not, the user should be redirected to the logout or /index.

justinwray:
@juanrossi 

Thank you for the report.  This has been fixed and will be seen in an upcoming PR.

juanrossi:
@justinwray Excellent, thanks for letting me know!

PR #565, opened 2017-09-03
Prevent extra frontend requests on errors

juanrossi:
Hello, having lots of users playing the game some times generates errors on the backend side, or some latency. 

Last year we had some issues because our server was overloaded. When a user sent a request it took too long, and the frontend sent new ones without waiting for the first one (creating more issues on the server).

This PR will prevent two things:

* If a request was sent, don't create another one until the first one has finished.
* If a request fails, wait one extra "round" before sending new ones.

All suggestions are welcome, I'm not a frontender so I'm sure this can be improved.

### DIFF
--- a/src/static/js/fb-ctf.js
+++ b/src/static/js/fb-ctf.js
@@ -274,7 +274,9 @@ function setupInputListeners() {
         $listview,
         $mapSvg,
         $map,
-        $countryHover;
+        $countryHover,
+        reload = true,
+        reload_team = true;
 
 
     /**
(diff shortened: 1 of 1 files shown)


justinwray:
@juanrossi Thank you for the contribution!  I am not opposed to the idea of blocking the AJAX requests to prevent error/failed requests from stacking up, and lowering additional load in the case of high latency.  There _might_ be a better way to handle the blocking; we will need to look into this.

Either way, we will take a look at what you have provided and let you know.

juanrossi:
@justinwray excellent! Yeah, for sure there are better ways for blocking this type of requests and prevent stacking.

This could also help with #563, to make sure users are redirected and not continue doing requests.

stevcoll:
@juanrossi Don't see any issues, @justinwray ready for your review.

justinwray:
@juanrossi After review and consideration of the method - we are going to **merge** this PR as is.  Though there are some other methods to do blocking AJAX requests, they come at different costs and this implementation provides the cleanest and most straightforward method given the desired goals.

A potential to-do:  rename the status/count variables to be more clear about their purpose.

Thank you again for your submission!

juanrossi:
Ok, I'm fine with that. What names do you propose?

justinwray:
@juanrossi 

Maybe `request_count` and `request_reload` - a few comments (when they are defined) may also be valuable.  We can make those updates in a separate PR.  I am going to merge this one now.

PR #573, opened 2017-09-28
Facebook and Google Login Integration & LiveSync Update

justinwray:
* The platform now supports Login via OAuth2 for Facebook and Google.  When configured and enabled, users will have the option to login to their existing account with a Facebook or Google account.

* Automated registration through Facebook or Google OAuth2 is now supported.  When configured and enabled, users will have the option to register an account by using and linking an existing account with Facebook or Google.

* Added "facebook_login" to the database schema.  This configuration option is a togglable setting to enable or disable login via Facebook.

* Added "google_login" to the database schema.  This configuration option is a togglable setting to enable or disable login via Google.

* Added "facebook_registration" to the database schema.  This configuration option is a togglable setting to enable or disable registration via Facebook.

* Added "google_registration" to the database schema.  This configuration option is a togglable setting to enable or disable registration via Google.

* Added "registration_prefix" to the database schema.  This configuration option is a string that sets the prefix for the randomly generated username/team name for teams registered via (Facebook or Google) OAuth.

* New Integration section within the Administrative interface allows for control over the Facebook and Google Login/Registration options and the automated team name prefix.

* Overhauled the Login page to support the new Login buttons to to display appropriate messages based on the configuration of login.

* Changed the "Sign Up" link to a button on the login page.

* Login form is dynamically generated, based on the configuration options and settings.

* Overhauled the Registration page to support the new Registration buttons to to display appropriate messages based on the configuration of registration.

* Registration form is dynamically generated, based on the configuration options and settings.

* Account Linking for Facebook now sets both the Login OAuth values and the LiveSync values (single step for both).

* Account Linking for Google now sets both the Login OAuth values and the LiveSync values (single step for both).

* Facebook Account linkage option has been added to the Account modal.

* The Account modal now shows which accounts are already linked.

* The Account modal will color-code the buttons on error (red) and success (green).

* Users can now change their team name from within the Account modal.

* New table "teams_oauth" has been added to handle the OAuth data for Facebook and Google account linkage.

* New class `Integration` handles the linkage of Facebook or Google accounts with an FBCTF account (both Login OAuth values and the LiveSync values).  The Integration class also includes the underlying methods for authentication in both the linkage and login routines and the oauth registration process.

* New data endpoint `data/integration_login.php`.  This endpoint accepts a type argument, currently supporting types of 'Facebook' and 'google'.  Through this endpoint, the login process is handled in conjunction with the Integration class.

  * The new callback URL for Facebook Login:  `/data/integration_login.php?type=facebook`

  * The new callback URL for Google Login:  `/data/integration_login.php?type=google`

* New data endpoint `data/integration_oauth.php`.  This endpoint accepts a type argument, currently supporting types of 'Facebook' and 'google'.  Through this endpoint, the oatuth account linkage is handled in conjunction with the Integration class.

  * The new callback URL for Facebook linkage:  `/data/integration_login.php?type=facebook`

  * The new callback URL for Google linkage:  `/data/integration_login.php?type=google`

* Old Google-specific endpoint (`data/google_oauth.php`) has been removed.

* New Team class methods:  genAuthTokenExists(), genTeamFromOAuthToken(), genSetOAuthToken().

* Team::genAuthTokenExists() allows an OAuth token to be verified.

* Team::genTeamFromOAuthToken() returns a Team object based on the OAuth token supplied.

* Team::genSetOAuthToken() sets the OAuth token for a team.

* Settings.ini and Configuration have methods to verify and return Facebook API settings.

* LiveSync has been updated to support and supply Facebook and Google OAuth output.  All of the users LiveSync integrations (FBCTF, Facebook, and Google) are now provided through the API.  As a result, so long as one of the three LiveSync methods are configured by the user (which happens automatically when linking an account to Facebook or Google) the data will become available through the LiveSync API.

* LiveSync now includes a "general" type.  The new "general" type output includes the scoring information using the local team name on the FBCTF instance.  This new type is not for importation on another FBCTF instance but does provide the opportunity for third-parties to use the data for score tracking, metric collections, and displays.  As such, this new LiveSync data allows the scoring data for a single FBCTF instance to be tracked.

* The `live_import.sh` script, used to import LiveSync API data will ignore the new "general" LiveSync type.

* Updated numerous Team::genTeam() calls to used the cached version:  MultiTeam::genTeam().

* Formatted all code files that were edited as part of this PR.

* In order to use the new Facebook or Google integration the following must be completed:

  * A Facebook and/or Google Application must be created, and OAuth2 API keys must be obtained.

  * The API keys must be provided in the Settings.ini file.

  * Desired settings must be configured from within the administrative interface (`Configuration`) - the default has all integration turned off.

* Note:  Facebook Login/Integration will _not_ work in `development` mode - this is due to a pending issue in the Facebook Graph SDK (facebook/php-graph-sdk#853) utilization of the pending PR (facebook/php-graph-sdk#854) resolves this issue.  Alternatively, the Integration with Facebook will work in `production` mode, the recommended mode for a live game.


src/controllers/ajax/IndexAjaxController.php
```diff
@@ -157,10 +158,12 @@ protected function getActions(): array<string> {
       // This will help avoid leaking users ldap passwords if the server's database
       // is compromised.
       $ldap_password = $password;
-      $password = strval(gmp_strval(
-        gmp_init(bin2hex(openssl_random_pseudo_bytes(16)), 16),
-        62,
-      ));
+      $password = strval(
+        gmp_strval(
+          gmp_init(bin2hex(openssl_random_pseudo_bytes(16)), 16),
```
fredemmott: while you're touching this, changing this to use `random_bytes(16)` would be good - and potentially using the raw bytes, hex, or base64 instead of requiring GMP to do a base62 conversion

src/models/Configuration.php
```diff
@@ -132,6 +133,39 @@ private static function configurationFromRow(
     );
   }
 
+  public static function genFacebookOAuthSettingsExists(): bool {
```
fredemmott: FBCTF is mostly following the FB standard of 'gen prefix means returns Awaitable'; google oauth is currently an exception, but shouldn't be.

src/models/Configuration.php
```diff
@@ -132,6 +133,39 @@ private static function configurationFromRow(
...
+    if ((array_key_exists('FACEBOOK_OAUTH_APP_ID', $config) === true) &&
+        (array_key_exists('FACEBOOK_OAUTH_APP_SECRET', $config) === true)) {
+      return true;
+    }
+    return false;
+  }
+
+  public static function genFacebookOAuthSettingsAppId(): string {
+    $settings_file = '../../settings.ini';
+    $config = parse_ini_file($settings_file);
```
fredemmott: For both this and the existing google things, we shouldn't have this repeated as often; split the ini parsing to a function, cache it, and use TypeAssert to assert it matches your expected safe to get typesafety

src/models/Integration.php
```diff
@@ -0,0 +1,379 @@
...
+  }
+
+  public static async function facebookLoginEnabled(): Awaitable<bool> {
+    $login_facebook = await Configuration::gen('login_facebook');
+    $oauth = Configuration::genFacebookOAuthSettingsExists();
+
+    $login_facebook_enabled =
+      $login_facebook->getValue() === '1' ? true : false;
+
+    if (($oauth) && ($login_facebook_enabled)) {
```
fredemmott: shouldn't generally add unneccessary brackets like this; the typechecker takes it as a "i know what I'm doing" hint in some situations, e.g. when assigning in an if condition

src/models/Integration.php
```diff
@@ -0,0 +1,379 @@
...
+    SessionUtils::sessionRefresh();
+    if (!SessionUtils::sessionActive()) {
+      SessionUtils::sessionSet('team_id', strval($team->getId()));
+      SessionUtils::sessionSet('name', $team->getName());
+      SessionUtils::sessionSet(
+        'csrf_token',
+        (string) gmp_strval(
+          gmp_init(bin2hex(openssl_random_pseudo_bytes(16)), 16),
+          62,
+        ),
```
fredemmott: - use random_bytes() instead of openssl_random_pseudo_bytes()
 - similar: why base62 with gmp instead of base64 or hex?

src/models/Integration.php
```diff
@@ -0,0 +1,379 @@
...
+        );
+        return $set_integrations;
+      }
+    } else if ($error !== false) {
+      return false;
+    }
+
+    header('Location: '.filter_var($url, FILTER_SANITIZE_URL));
+    exit;
+    return false;
```
fredemmott: Is this required? The typechecker should realize it's unreachable

src/models/Integration.php
```diff
@@ -0,0 +1,379 @@
...
+    int $team_id,
+    string $type,
+    string $email,
+    string $id,
+  ): Awaitable<bool> {
+    $livesync_password_update =
+      await Team::genSetLiveSyncPassword($team_id, $type, $email, $id);
+
+    $oauth_token_update =
+      await Team::genSetOAuthToken($team_id, $type, $email);
```
fredemmott: when independent:

```Hack
list($a, $b) = await \HH\Asio\va(foo::genA(), foo::genB());
```

Otherwise async just adds overhead, and can't get you anything.

src/models/Integration.php
```diff
@@ -0,0 +1,379 @@
...
+      return false;
+    }
+  }
+
+  public static async function genRegisterTeam(
+    string $email,
+    string $id,
+  ): Awaitable<int> {
+    $registration_prefix = await Configuration::gen('registration_prefix');
+    $logo_name = await Logo::genRandomLogo();
```
fredemmott: ditto

src/models/Team.php
```diff
@@ -408,6 +408,38 @@ public static function regenerateHash(string $password_hash): bool {
...
+      $team = await self::genTeam($team_id);
+      await self::genUpdate(
+        $shortname,
+        $team->getLogo(),
+        $team->getPoints(),
+        $team_id,
+      );
+      MultiTeam::invalidateMCRecords(); // Invalidate Memcached MultiTeam data.
+      ScoreLog::invalidateMCRecords(); // Invalidate Memcached ScoreLog data.
+      ActivityLog::invalidateMCRecords(); // Invalidate Memcached ActivityLog data.
```
fredemmott: The comments don't say anything the code doesn't

src/models/Team.php
```diff
@@ -734,29 +766,146 @@ public static function regenerateHash(string $password_hash): bool {
...
+  ): Awaitable<bool> {
+    $db = await self::genDb();
+
+    $team_id_result = await $db->queryf(
+      'SELECT id FROM teams_oauth WHERE type = %s AND team_id = %d',
+      $type,
+      $team_id,
+    );
+
+    if ($team_id_result->numRows() === 1) {
```
fredemmott: style nit: we'd generally write:

```
return ($team_id_result->numRows() === 1);
```

Also, when if blocks return, we tend to not do an else block, to reduce nesting, and make it clearer what the flow is:

```
if (foo) {
  return true;
}
return false;
```

This applies throughout your changes.

src/scripts/liveimport.php
```diff
@@ -362,6 +362,10 @@ function($a, $b) {
     $teams_array = array();
     $teams = await self::genTeamDefaults($teams);
     foreach ($teams as $team_livesync_key => $team_data) {
+      list($type, $username, $key) = explode(':', $team_livesync_key);
+      if ($type === 'general') {
```
fredemmott: should there be an enum or const for valid values of $type?

src/controllers/modals/ActionModalController.php
```diff
@@ -266,20 +266,83 @@ class="fb-cta cta--yellow js-close-modal js-confirm-save">
           <h4>
             {tr('account_')}<span class="highlighted">{tr('Settings')}</span>
           </h4>;
+        $oauth_header = '';
+        if (Configuration::genFacebookOAuthSettingsExists() === true) {
+          $linked = \HH\Asio\join(
```
fredemmott: Not in this PR, but in general, there should be one call to `\HH\Asio\join()` per HTTP request; that'll need some refactoring of the model classes.

src/controllers/ajax/IndexAjaxController.php
justinwray: Good idea.  We can swap this out in a few places while we are at it.  Currently `openssl_random_pseudo_bytes()` is in use in:

* IndexAjaxController (twice).
* Integration
* Token.

src/models/Configuration.php
justinwray: Not a problem.  We will update this.

src/models/Configuration.php
justinwray: Great idea.  We can certainly cache this and not repeatedly parse the API keys from disk.

src/models/Integration.php
justinwray: We'll fix that here.  I wonder where else we may be using extra brackets within the platform.  We should take a look at this soon.

src/models/Integration.php
justinwray: Seems to be.  We were getting an error from the typechecker, and we didn't want to make the method nullable.

src/models/Integration.php
justinwray: Absolutely!  We normally combine the awaitables and intend to do so in this PR.

src/models/Team.php
justinwray: This is another one that would be good to fix throughout the code base.

src/scripts/liveimport.php
justinwray: Maybe.  Part of the idea here is to support the import from other platforms (not exclusively FBCTF platforms).  For that reason we actually provided fallthroughs (`default`) in the `switch` statements in the various related classes.

In this file specifically, we just want to ignore the `general` values which is unique to the LiveSync API (it allows external platforms to get the scoring data with the team names from the platform, but is not useful for importing).

src/controllers/modals/ActionModalController.php
justinwray: +1 on this!

justinwray:
@fredemmott Thank you for the feedback!  We will be updating this PR soon with the changes discussed.

justinwray:

PR #574, opened 2017-09-28
Local Caching of Database and Cache Results

justinwray:
* Results from the database (MySQL) are stored and queried from the Cache service (Memcached).  Results from Memcached are now stored locally within HHVM memory as well.

* New class Cache has been added, with three methods:

  * setCache()
  * getCache()
  * deleteCache()

* Through the new Cache class a global cache object is created and used throughout a single HHVM execution thread (user request).

* This change results in a 99.9995% reduction in Memcached requests (at scale), providing a major improvement on scalability and performance.

### DIFF
--- a/src/models/Model.php
+++ b/src/models/Model.php
@@ -1,6 +1,7 @@
 <?hh // strict
 
 abstract class Model {
+
   protected static Db $db = MUST_MODIFY;
   protected static Memcached $mc = MUST_MODIFY;
   protected static string $MC_KEY = MUST_MODIFY;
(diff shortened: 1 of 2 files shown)


fredemmott:
Static member variables would be a good alternative to $GLOBALS; $GLOBALS and variables in the global scope can not be typed.

Also, the $GLOBAL_CACHE property seems mis-named - it's just a standard instance property, in no way global.

src/models/Model.php
```diff
@@ -1,6 +1,7 @@
 <?hh // strict
 
 abstract class Model {
+
```
fredemmott: could either make Cache a singleton, or add a protected static Cache (or just the Map/dict and remove the Cache class) here

justinwray:

PR #575, opened 2017-09-28
Blocking AJAX Requests

justinwray:
* AJAX requests for the gameboard are now individually blocking.  A new request will not be dispatched until the previous request has completed.

* AJAX requests will individually stop subsequent requests on a hard-error.


justinwray:

PR #590, opened 2017-10-24
Attachment Security Update

justinwray:
* Attachments have been moved out of a web accessible direcotry (`/var/www/fbctf/src/data/`) and now reside in `/var/www/fbctf/attachments`.

* Attachment downloads are now handled by an endpoint in `/data`: `attachment.php`.

* All links to attachments now refer to the correct `attachment.php` endpoint location.

* The `tar` command within the Attachment Import function now sets the mode to 600 at time of extraction.

* The Attachment Import functionality will no longer change or attempt to change permissions on the current or parent directories.

* The Attachment Import functionality will no longer change permissions on any directories, though subdirectories are not supported.

* Attachment filenames will no longer be altered, excluding the inclusion of the file hash.

* The provision script has been updated to support the new Attachment directory location.

* Attachment specific directives are no longer set in the Mult-Server Nginx configuration.

* Attachment location information has been updated in the `.gitignore` configuration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,10 +31,10 @@ www.pid
 *.swo
 
 # Attachments directory
-src/data/attachments/*
-!src/data/attachments/index.php
-src/data/attachments/deleted/*
-!src/data/attachments/deleted/index.php
+attachments/*
+!attachments/index.php
+attachments/deleted/*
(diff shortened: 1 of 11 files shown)


javuto:
17pmxu

PR #591, opened 2017-10-24
Google OAuth Security Update

justinwray:
* The Google OAuth process now sets a CSRF token.

### DIFF
--- a/src/data/google_oauth.php
+++ b/src/data/google_oauth.php
@@ -22,7 +22,16 @@ if (Configuration::genGoogleOAuthFileExists()) {
   );
 
   $integration_csrf_token = base64_encode(random_bytes(100));
-  setcookie('integration_csrf_token', strval($integration_csrf_token));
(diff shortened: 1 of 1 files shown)


src/data/google_oauth.php
```diff
@@ -16,8 +21,23 @@
     'https://'.$_SERVER['HTTP_HOST'].'/data/google_oauth.php',
   );
 
-  if (isset($_GET['code'])) {
-    $client->authenticate($_GET['code']);
+  $integration_csrf_token = base64_encode(random_bytes(100));
+  setcookie('integration_csrf_token', strval($integration_csrf_token));
```
gsingh93: Need to set the secure/httponly flag. Also, add a comment mentioning the fact that the cookie isn't actually set until the request is sent down, so we aren't actually clobbering the cookie we use below (this might be obvious to most PHP developers, but not to me).

javuto:
shiptit_letitbegin

PR #592, opened 2017-10-26
Changing teamname into team_name

asudhak:
Bug in signing up new users due to a typo in the teamname element. Related to https://github.com/facebook/fbctf/commit/1a8286bef5d124e85e48bb281a00cfd5477152f2

Without this, verifyTeamname (https://github.com/asudhak/fbctf/blob/dev/src/static/js/index.js#L46) during user sign up will fail because there is not element called 'team_name'

### DIFF
--- a/src/controllers/IndexController.php
+++ b/src/controllers/IndexController.php
@@ -347,7 +347,7 @@ class IndexController extends Controller {
                 <label for="">{tr('Team Name')}</label>
                 <input
                   autocomplete="off"
-                  name="teamname"
+                  name="team_name"
                   type="text"
                   maxlength={20}
                   autofocus={true}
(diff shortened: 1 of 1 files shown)


justinwray:
@asudhak 

Thank you for your contribution.

`verifyTeamName()` uses `teamname` see:  https://github.com/facebook/fbctf/blob/master/src/static/js/index.js#L44

There does seem to be some inconsistencies, like https://github.com/facebook/fbctf/blob/master/src/static/js/index.js#L206 as well.

This will need to be corrected in multiple places and match on both the PHP and JS side (including the input filters).

asudhak:
@justinwray the pull request is only for the dev branch: https://github.com/facebook/fbctf/blob/dev/src/static/js/index.js

Seems like there was some work done to rename stuff from teamname to team_name, and there were only 2 places in the dev branch that got missed.

People who use the dev branch cannot 'sign up/register' without the above fix.

justinwray:
Thank you again for your submission!

The utilization of the team name variable was reviewed, and corrected, across the PHP/JS code in #594 and merged into `master` in #605.

If you notice other issues with login, or the team name variable, please open a new issue or submit a new pull request.

PR #612, opened 2018-01-08
Fixes #588: Level Import: Random ISO Code

### LINKED ISSUE #588: Level Import: Random ISO Code
I'm trying to batch import levels into my instance, while I don't have an ISO code for each one, is there a way to randomly assign an ISO code to each level like the Admin Interface allows (Selecting Auto in the drop down)?

(Commit 1f236bb Production on 16.04)

mptap:


src/models/Level.php
```diff
@@ -151,7 +151,7 @@ private static function levelFromRow(Map<string, string> $row): Level {
     foreach ($elements as $level) {
       $title = must_have_string($level, 'title');
       $type = must_have_string($level, 'type');
-      $entity_iso_code = must_have_string($level, 'entity_iso_code');
+      $entity_iso_code = must_have_string($level, 'entity_iso_code') or await Country::genRandomAvailableCountry()->getIsoCode();
```
azjezz, line 154: revert.
1. always use `||` instead of `or`
2. this is a syntax error, we can't have `await` in there.
3. `must_have_string` will always return, there's no case where it returns a "falsy" value, so the `or await Country::genRandomAvailableCountry()->getIsoCode();` will never be executed.

Issue #617, opened 2018-01-29
Attachments corrupted when downloading


jnwarp:
When downloading a file, the file appears broken on certain files.  This is because the file size is being appended to the file when downloading.

__Steps to reproduce:__
1. Upload an attachment as a text file
2. Download attachment through challenge
3. Look at the downloaded file, there is a number appended.

__Screenshot:__
<img src="https://i.imgur.com/5Wgcktt.png">

__Problematic code:__
*src/data/attachments.php*
```php
$data = readfile(Attachment::attachmentsDir.$filename);
```

The read file function appends the length to the end of the file. Instead it should be replaced with `file_get_contents`.

hahnakane:
I have this same issue as well.

jnwarp:
@gsingh93 would you mind taking a look at this issue? I'll be deploying an instance of FBCTF soon, it would really help to make sure that the fix is in one of the branches.

Please let me know if there are any adjustments you'd like me to make.

gsingh93:
Merged that PR, thanks!

PR #618, opened 2018-01-29
Fix attachments code

jnwarp:
When downloading an attachment, the byte length of the file is appended, see issue #617.


gsingh93:
Thanks! What's actually happening here is that `readfile` is writing to the output buffer directly and the return value is the number of bytes read, which is being stored into `$data`. `$data` then gets sent down in the base class (`DataController`). While `readfile` is the more efficient way to do things, I think `file_get_contents` is cleaner in this case.

Issue #113, opened 2016-05-30
Countdown timer to next game


swappage:
Hello,
i'm having an hard time in figuring out how to properly set up the countdown page.

I can't seem to find a configuration in the admin panel, but also by editing the value in the configuration table of the database doesn't seem to affect the page behaviour.

It always says "SOON"

I took a quick look at the controller int he source and the genRenderCountdownContent() should render the next date event if the value is set in the database or SOON if it's not set.
Maybe i'm using the wrong format for the value? the description of the parameter in the DB says it's a date so i used the format
"YYYY-MM-DD hh:mm:ss"

also.. is the countdown timer routine implemented at all? in the index page?

```html        
<li><span class="count-number">--</span>_days</li>
<li><span class="count-number">--</span>_hours</li>
 <li><span class="count-number">--</span>_minutes</li>
<li><span class="count-number">--</span>_seconds</li>
```

thanks for the help.

hackmethod:
@swappage Did you get a response on this matter?  I would also like to know how I can set a countdown timer for the next CTF we host.

gsingh93:
Sorry for not responding, we labelled the issue meaning it's something we need to fix. We don't have an ETA on a fix at the moment.

leobraga96:
@swappage Did you get a response on this matter? I would also like to know how I can set a countdown timer for the next CTF

PR #630, opened 2018-03-14
Make game start countdown timer count down

JasonTarka:
Rather than being a static number updated on page refresh, the
countdown clock will not update itself. This makes for a more
interesting display.

Using time remaining rather than an absolute time to prevent
jittery behaviour at the start, though at the expense of accuracy.


JasonTarka:
This should fix #113

PR #635, opened 2018-03-27
Base improvements

JasonTarka:
Improve appearance of bases & links
- Show points for both capturing & holding a base
- Show non-http(s) links as full-width, so they're easier to read
- Show "Capture Points" & "Keep points" when editing bases instead of "Points" & "Bonus"

Mark bases as "held" when the team doesn't change
- Mark as 'held' rather than 'captured' when the controlling team hasn't changed
- Don't spam the gameboard's activity log with messages of a team holding a base
- Show in the activity log when a base is captured by another team

Reverse points/bonus for bases
- More logical to have main points be the capture points, and bonus be the hold points, rather than points = hold, bonus = capture.
- Adding version detection to allow old exports to be imported easily.
- Extra logging to make debugging easier/possible

Functional fixes for bases & links
- Handle http/https/ftp/etc. prefix for base URLs
  - Previously `http://example.com` would lead to the system calling `http:12345`; will now call example.com:12345
- Better display of non-http(s) links
  - If there's a port listed, show the full `nc 1.2.3.4 567`, if not, show just `1.2.3.4` (previously would show `nc 1.2.3.4 undefined)
  - Readonly, rather than disabled, so the text can be scrolled & selected

### DIFF
--- a/src/controllers/AdminController.php
+++ b/src/controllers/AdminController.php
@@ -2540,13 +2540,13 @@ class AdminController extends Controller {
                   <div
                     class=
                       "form-el--required col col-1-2 el--block-label el--full-text">
-                    <label>{tr('Keep Points')}</label>
+                    <label>{tr('Capture Points')}</label>
                     <input name="points" type="text" />
                   </div>
                   <div
                     class=
                       "form-el--required col col-1-2 el--block-label el--full-text">
-                    <label>{tr('Capture points')}</label>
+                    <label>{tr('Keep points')}</label>
                     <input name="bonus" type="text" />
                   </div>
                 </div>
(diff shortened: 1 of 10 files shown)


JasonTarka:
Note: This relies on #634 due to using `Utils::logMessage()`. Using a separate PR to make it easier to review.

icebowl:
Please throw us a bone.
Last year we ran fbctf and got the base to work.  This year we have a problem.
fbctf installed on base metal Ubuntu.
fbctf 10.183.7.7/16   base 10.183.7.31/16  .   Then base and the game are on the same level 1 switch.
We can not get the base to work.
Base url 10.183,7.31:12345
STEPS:
- We create a jail chroot.
- Run pythons sciprt on for port 12345
- Start the game then create the BASE.
- No luck.

PR #636, opened 2018-04-04
Fix letsencrypt certbot script

3553x:
Previous arguments to cerbot-auto are deprecated and appear to be unsupported.

The setup script `extra/provision.sh` fails and outputs the following messages when it's run with `-c certbot` as an argument:
`Client with the currently selected authenticator does not support any combination of challenges that will satisfy the CA.`
`The standalone specific supported challenges flag is deprecated. Please use the --preferred-challenges flag instead.`
There was also a message about rerunning certbot as root which I didn't capture.

Also see:
https://community.letsencrypt.org/t/important-what-you-need-to-know-about-tls-sni-validation-issues/50811
Which states that `tls-sni` is not supported at the moment and
https://letsencrypt.readthedocs.io/en/latest/using.html#standalone
which states that `--standalone-supported-challenges` is deprecated.

### DIFF
--- a/extra/lib.sh
+++ b/extra/lib.sh
@@ -132,14 +132,14 @@ function letsencrypt_cert() {
     cat <<- EOF > /root/tmp/certbot.sh
 		#!/bin/bash
 		if [[ ! ( -d /etc/letsencrypt && "\$(ls -A /etc/letsencrypt)" ) ]]; then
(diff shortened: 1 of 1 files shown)


vegardvaage:
Just tested for a standalone install, works great. Thanks!

